PR #10, opened 2023-08-11
Initial ORM implementation

pking-moj:
## Link to relevant ticket

[CRM457-](https://dsdmoj.atlassian.net/browse/CRM457-)

### DIFF
--- a/alembic/env.py
+++ b/alembic/env.py
@@ -64,7 +64,7 @@ def run_migrations_online() -> None:
             os.getenv("POSTGRES_USERNAME", "test"),
             os.getenv("POSTGRES_PASSWORD", "pass"),
             os.getenv("POSTGRES_HOSTNAME", "localhost"),
-            os.getenv("POSTGRES_dbNAME", "laa_crime_application_store"),
+            os.getenv("POSTGRES_DBNAME", "laa_crime_application_store"),
         )
 
         config.set_main_option("sqlalchemy.url", postgres_url)
--- a/alembic/env.py
+++ b/alembic/env.py
@@ -81,9 +81,7 @@ def run_migrations_online() -> None:
             with context.begin_transaction():
                 context.run_migrations()
     else:
-        context.configure(
-            connection=connectable, target_metadata=target_metadata
-        )
+        context.configure(connection=connectable, target_metadata=target_metadata)
 
         with context.begin_transaction():
             context.run_migrations()
--- a/alembic/env.py
+++ b/alembic/env.py
@@ -57,23 +57,33 @@ def run_migrations_online() -> None:
     and associate a connection with the context.
 
     """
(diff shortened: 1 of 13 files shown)


alembic/env.py
```diff
@@ -57,23 +57,31 @@ def run_migrations_online() -> None:
...
-    with connectable.connect() as connection:
-        context.configure(connection=connection, target_metadata=target_metadata)
+    connectable = config.attributes.get("connection", None)
+
+    if connectable is None:
+        postgres_url = "postgresql+psycopg2://{}:{}@{}/{}".format(
+            os.getenv("POSTGRES_USERNAME", "test"),
+            os.getenv("POSTGRES_PASSWORD", "pass"),
+            os.getenv("POSTGRES_HOSTNAME", "localhost"),
+            os.getenv("POSTGRES_dbNAME", "laa_crime_application_store"),
```
dwhenry, line 67: lower case db?

tests/data/database_test.py
```diff
@@ -0,0 +1,9 @@
+import os
+
+from laa_crime_application_store_app.data.database import get_db
+
+
+def test_get_db_returns_valid_session():
+    os.environ["POSTGRES_NAME"] = "laa_crime_application_store_test"
```
dwhenry, line 7: it feels like we have multiple approaches to the name setting.

Here we are hardcoding it, but in the migrations we appended to the existing name, I feel this is slightly confusing and would prefer a consistent approach if possible. :)

dwhenry (commented):
some small comments

alembic/env.py
pking-moj: fixed

tests/data/database_test.py
pking-moj: Have normalised how we do this

dwhenry (approved):

laa_crime_application_store_app/routers/application.py
```diff
@@ -0,0 +1,25 @@
...
+@router.get("/application/{app_id}", response_model=App)
+async def ping(app_id: UUID | None = None, db: Session = Depends(get_db)):
+    logger.info("GETTING_APPLICATION", application_id=app_id)
+    application = db.query(Application).filter(Application.id == app_id).first()
+    if application is not None:
+        logger.info("APPLICATION_FOUND", application_id=app_id)
+        return App(**application.__dict__)
+
+    logger.info("APPLICATION_NOT_FOUND", application_id=app_id)
+    return Response(status_code=400)
```
ivanELEC, line 25: is the name of the method for this endpoint still meant to be ping? Seems to be an almost complete endpoint for our get requests (obviously more will be added to get versions etc)

ivanELEC (approved):

PR #12, opened 2023-08-16
Add put endpoint

dwhenry:
## Description of change
Add `PUT /application/{id}` endpoint

## Link to relevant ticket

[CRM457-388](https://dsdmoj.atlassian.net/browse/CRM457-388)

### DIFF
--- a/laa_crime_application_store_app/routers/application.py
+++ b/laa_crime_application_store_app/routers/application.py
@@ -18,7 +18,7 @@ router = APIRouter()
 logger = structlog.getLogger(__name__)
 
 responses = {
-    201: {"description": "Application has been created"},
+    201: {"description": "Application/Version has been created"},
     400: {"description": "Resource not found"},
     409: {"description": "Resource already exists"},
 }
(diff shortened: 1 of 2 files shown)


laa_crime_application_store_app/routers/application.py
```diff
@@ -83,3 +83,31 @@ async def post_application(request: ApplicationNew, db: Session = Depends(get_db
...
+
+
+@router.put("/application/{app_id}", status_code=201, responses=responses)
+async def put_application(
+    app_id: UUID, request: ApplicationNew, db: Session = Depends(get_db)
+):
+    application = db.query(Application).filter(Application.id == app_id).first()
+    if application is None:
+        logger.info("APPLICATION_NOT_FOUND", application_id=app_id)
+        return Response(status_code=400)
```
pking-moj, line 95: If there is no application, this should be considered and upsert and therefor create a new entry - https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/PUT

dwhenry: then what is the point of the post endpoint...

laa_crime_application_store_app/routers/application.py
```diff
@@ -83,3 +85,51 @@ async def post_application(request: ApplicationNew, db: Session = Depends(get_db
...
+
+    # we ignore the issue if no application version is found to
+    # avoid the system getting into a state where new versions
+    # can't be added.
+    if (
+        application_version
+        and application_version.application == request.application
+        and application.application_state == request.application_state
+    ):
+        return Response(status_code=200)
```
pking-moj, line 116: This should probably be a 204 - No Content

laa_crime_application_store_app/routers/application.py
pking-moj: This code is still reading as if there isnt an application then we are returning a 400. The update should still have the risk value as it can be amended by the caseworker.

laa_crime_application_store_app/routers/application.py
dwhenry: updated

pking-moj (approved):

PR #20, opened 2023-10-10
Add application type to store

pking-moj:
## Description of change

- Adds application_type to the DB and updates existing rows to be crm7
- Adds application_type to models to allow it to come through in requests
- Updates tests

## Link to relevant ticket

N/A

### DIFF
--- a/alembic/versions/3a99fcfd6934_add_form_type.py
+++ b/alembic/versions/3a99fcfd6934_add_form_type.py
@@ -19,18 +19,13 @@ depends_on: Union[str, Sequence[str], None] = None
 
 
 def upgrade() -> None:
-    op.add_column("application", sa.Column("application_type", sa.Text, nullable=False))
+    op.add_column("application", sa.Column("application_type", sa.Text))
 
-    rows = sa.text("select* from application")
-
+    update_rows = sa.text("update application set application_type = 'crm7'")
     connection = op.get_bind()
+    connection.execute(update_rows)
 
-    for row in connection.execute(rows):
-        connection.execute(
-            sa.text(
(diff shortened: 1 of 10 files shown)


imtiazAhmd (approved):

alembic/versions/3a99fcfd6934_add_form_type.py
```diff
@@ -0,0 +1,37 @@
...
+    op.add_column("application", sa.Column("application_type", sa.Text, nullable=False))
+
+    rows = sa.text("select* from application")
+
+    connection = op.get_bind()
+
+    for row in connection.execute(rows):
+        connection.execute(
+            sa.text(
+                "update application set application_type = crm7 where id = %s" % row.id
```
dwhenry, line 31: have you run this... as shouldn't it be
```
"update application set application_type = 'crm7' where id = %s" % row.id
```

but more than that, any reason you didn't just do a single update on all rows?

dwhenry (commented):
Unsure about the migration, otherwise looks good.. 

Also this is a breaking change - the other option would be to have the application_type default to crm7 if it is not set... that would avoid a breaking chnage and we could remove the default once we updated the other apps... I guess we're not live so not really required right now...

alembic/versions/3a99fcfd6934_add_form_type.py
pking-moj: Updated to something workable and have tested. About the default - i want to try and avoid the assumptions in this service hence not going that route. Before live we can wipe the data and start fresh to ensure integrity if needed

dwhenry (approved):

PR #88, opened 2024-02-21
Add documentation about env vars

jsugarman:
## Description of change
Add documentation and sample for expected env vars

Not having the APP_CLIENT_ID, TENANT_ID
secrets will cause the server to timeout
locally

Not having the SWAGGER_ENDPOINT var
means you will not be able to get to
the swagger docs `/docs` locally.

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -1,6 +1,12 @@
+# Database setup requirements
 POSTGRES_USERNAME=
 POSTGRES_PASSWORD=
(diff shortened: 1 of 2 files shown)


ivanELEC (approved):

ivanELEC:
strange, the --upgrade command on the apk add for expat in the dockerfile should automatically upgrade it to the newest version which doesn't have this security vulnerability - going to re-run the pipeline for you and see if this fixes it

PR #100, opened 2024-03-12
Use correct namespace for ingress

ivanELEC:
## Description of change
Use “laa-crime-application-store-production”  instead of “laa-crime-application-store-prod” for production ingress namespace


jogilder (approved):
LGTM

PR #107, opened 2024-03-22
Crm457 1146/permissions

dwhenry:
## Description of change
Add permission to updating the application

## Link to relevant ticket

[CRM457-1145](https://dsdmoj.atlassian.net/browse/CRM457-1145)

[CRM457-1145]: https://dsdmoj.atlassian.net/browse/CRM457-1145?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


laa_crime_application_store_app/services/permissions.py
```diff
@@ -0,0 +1,39 @@
+from typing import List
+
+from sentry_sdk import capture_message
+
+from laa_crime_application_store_app.models.application_schema import Application
+
+
+class Permissions:
+    LOCKED_STATES = ("granted", "expired", "auto_grant")
```
patrick-laa, line 9: I think CRM7 states `further_info` and `provider_requested` are also locked (or maybe they're provider-editable is we think the provider could in theory amend and re-submit them in the future)

laa_crime_application_store_app/services/permissions.py
```diff
@@ -0,0 +1,39 @@
...
+    CASEWORKER_EDITABLE_STATES = ("submitted", "provider_updated")
+    PROVIDER_EDITABLE_STATES = ("part_granted", "rejected", "part_grant", "sent_back")
```
patrick-laa, line 11: We've killed off `part_granted` now. It shouldn't exist anywhere.

laa_crime_application_store_app/services/permissions.py
```diff
@@ -0,0 +1,39 @@
+from typing import List
+
+from sentry_sdk import capture_message
+
+from laa_crime_application_store_app.models.application_schema import Application
+
+
+class Permissions:
+    LOCKED_STATES = ("granted", "expired", "auto_grant")
+    CASEWORKER_EDITABLE_STATES = ("submitted", "provider_updated")
```
patrick-laa, line 10: This needs to include `sent_back` because the caseworker app is responsible for updating `sent_back` applications to `expired` via a daily scheduled job.

laa_crime_application_store_app/services/v1/application_service.py
```diff
@@ -123,6 +125,9 @@ def update_existing_application(
         ):
             return existing_application.id
 
+        if not get_permissions().allow_update(existing_application, roles):
```
patrick-laa, line 128: We should have some sort of locking here. At the moment, it's not impossible that the provider app would try to update a `sent_back` PA application at the precise moment the caseworker app would try to mark it as expired. We don't want them both to pass the permission check simultaneously and then both apply their updates to the same record.

laa_crime_application_store_app/services/permissions.py
dwhenry: so this break out rule of only one app can edit in each state... :(

laa_crime_application_store_app/services/v1/application_service.py
dwhenry: So not going to worry about this in this PR as due to one being an automated task and the other manual submission there is low likelihood of overlap with submission times - if we are worries I would suggest changing the expiry task from 11pm to 4am.

patrick-laa: Is it not like a 2-line addition to add a 'SELECT FOR UPDATE'? The point of these PRs is to protect data integrity, and I'd rather our strategy not be "hope a lawyer doesn't pull an all-nighter"

laa_crime_application_store_app/services/permissions.py
```diff
@@ -0,0 +1,47 @@
...
+
+from laa_crime_application_store_app.config.permission_settings import get_permissions
+from laa_crime_application_store_app.models.application_schema import Application
+from laa_crime_application_store_app.services.auth_service import azure_schema
+
+permissins = get_permissions()
+
+
+def validate_can_create(user: User = Depends(azure_schema)) -> None:
+    if user is None or not user.roles:
```
jogilder, line 14: Didn't we say that we would have a check for whether or not the authentication was set here?
`settings.authentication_required.lower() != "true"`

jogilder: also would be better to not have this as an `elif`, can we just return from this if block and then have the next block as another if?

jogilder: And we won't remove this completely when the roles are implemented as we will still need to check if we have authentication turned on or not.

laa_crime_application_store_app/services/permissions.py
```diff
@@ -0,0 +1,47 @@
...
+    elif permissins.provider_role not in user.roles:
+        raise InvalidAuth("User not permitted to create application")
+
+
+def validate_can_update(
+    application: Application, user: User = Depends(azure_schema)
+) -> None:
+    if application.application_state in permissins.locked:
+        raise InvalidAuth("Application in locked state")
+    elif not user.roles:
```
jogilder, line 27: won't this fail if you have authentication turned off?

laa_crime_application_store_app/services/permissions.py
```diff
@@ -0,0 +1,47 @@
...
+        # remove this branch once roles have been implemented
+        # on Azure accounts
+    elif permissins.provider_role not in user.roles:
+        raise InvalidAuth("User not permitted to create application")
+
+
+def validate_can_update(
+    application: Application, user: User = Depends(azure_schema)
+) -> None:
+    if application.application_state in permissins.locked:
```
jogilder, line 25: typo in permissins

laa_crime_application_store_app/services/permissions.py
jogilder: Also if you wanted to we can use a match statement here as allowed from python 3.10 :)
https://docs.python.org/3/whatsnew/3.10.html#pep-634-structural-pattern-matching

laa_crime_application_store_app/config/permission_settings.py
```diff
@@ -0,0 +1,24 @@
...
+from pydantic_settings_yaml import YamlBaseSettings
+
+THIS_DIR = Path(__file__).parent
+
+
+class PermissionSettings(YamlBaseSettings):
+    model_config = SettingsConfigDict(yaml_file=f"{THIS_DIR}/permissions.yaml")
+
+    locked: List[str]
+    casework_editable: List[str]
```
jogilder, line 15:
```suggestion
    caseworker_editable: List[str]
```

laa_crime_application_store_app/config/permission_settings.py
```diff
@@ -0,0 +1,24 @@
...
+    provider_editable: List[str]
+    all_editable: List[str]
+    casework_role: str
```
jogilder, line 18:
```suggestion
    caseworker_role: str
```

laa_crime_application_store_app/services/permissions.py
```diff
@@ -0,0 +1,47 @@
+from fastapi import Depends
+from fastapi_azure_auth.exceptions import InvalidAuth
+from fastapi_azure_auth.user import User
+from sentry_sdk import capture_message
+
+from laa_crime_application_store_app.config.permission_settings import get_permissions
+from laa_crime_application_store_app.models.application_schema import Application
+from laa_crime_application_store_app.services.auth_service import azure_schema
+
+permissins = get_permissions()
```
jogilder, line 10:
```suggestion
permissions = get_permissions()
```

laa_crime_application_store_app/config/permissions.yaml
```diff
@@ -0,0 +1,16 @@
+locked:
```
jogilder, line 1: Have we got all states here?

laa_crime_application_store_app/internal/notifier.py
jogilder, line 1: can you document why this is removed in the conversation? I assume it is not used.

laa_crime_application_store_app/services/v1/application_service.py
```diff
@@ -168,7 +177,12 @@ def update_existing_application(
 
     @staticmethod
     def __get_application_by_id(db: Session, app_id: UUID):
-        return db.query(Application).filter(Application.id == app_id).first()
+        return (
+            db.query(Application)
+            .filter(Application.id == app_id)
+            .with_for_update()
```
jogilder, line 183: can you explain that this is added for row locking? And what error do we get if someone tries to edit at the same time and how is it captured and dealt with?

jogilder: There should be a test for race condition...

tests/routers/v1/application_test.py
```diff
@@ -88,11 +87,7 @@ def test_data_selected_version_returns_400(client: TestClient, seed_application)
     assert response.status_code == 400
 
 
-@patch("laa_crime_application_store_app.internal.notifier.Notifier.notify")
-def test_post_application_returns_200(
-    mock_notify, client: TestClient, dbsession: Session
-):
-    mock_notify.return_value = True
+def test_post_application_returns_200(client: TestClient, dbsession: Session):
```
jogilder, line 90: this is tricky because to my mind this test should fail as you have no roles (roles is an empty list) - The test should pass with a provider role and fail with caseworker or anything else as here we are mocking what happens when there is an authenticated user.
The issue is that we have methods that allow us to continue if there are no roles.

jogilder: We may have to feature flag this.

tests/services/conftest.py
```diff
@@ -0,0 +1,29 @@
+import pytest
+from fastapi_azure_auth.user import User
+
+
+def build_user():
```
jogilder, line 5: would be good to name this something different to that in the other conftest but not essential.

tests/services/permissions_test.py
```diff
@@ -0,0 +1,109 @@
...
+from fastapi_azure_auth.exceptions import InvalidAuth
+
+from laa_crime_application_store_app.models.application_schema import Application
+from laa_crime_application_store_app.services.permissions import (
+    validate_can_create,
+    validate_can_update,
+)
+
+
+def test_can_create_when_no_roles(normal_user):
```
jogilder, line 14: I don't think this test should be in here and maybe we feature flag this work instead.

tests/services/permissions_test.py
```diff
@@ -0,0 +1,109 @@
...
+    assert e.type == InvalidAuth
+    assert str(e.value) == "401: User not permitted to create application"
+
+
+def test_can_create_when_provider_role(normal_user):
+    normal_user.roles = ["Provider"]
+    assert validate_can_create(normal_user) is None
+
+
+def test_can_update_when_no_roles(normal_user):
```
jogilder, line 32: ditto this test

jogilder (commented):
@dwhenry sorry lots of comments.

patrick-laa:
Closing as we've done this in Ruby instead

PR #112, opened 2024-04-05
CRM457-1305: Skeleton for new notification service

patrick-laa:
## Description of change
We want to move towards immediate syncing rather than a 10-minute delay. To that end, the plan is that whenever a record is changed in the app store, all subscribers can be notified of the change. This provides the basic mechanism for that.

Still to do:
* Move webhook calls to an async Celery task so that they can be retried (with appropriate backoff) if they fail
* Only notify subscribers where the subscriber type is different to the role of the client initiating the change

## Link to relevant ticket

[CRM457-1305](https://dsdmoj.atlassian.net/browse/CRM457-1305)

[CRM457-1305]: https://dsdmoj.atlassian.net/browse/CRM457-1305?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


laa_crime_application_store_app/services/v1/notification_service.py
```diff
@@ -0,0 +1,64 @@
...
+            return existing_subscriber
+
+        new_subscriber = Subscriber(
+            subscriber_type=subscriber_type,
+            webhook_url=webhook_url,
+        )
+        db.add(new_subscriber)
+        db.commit()
+
+        return None
```
jogilder, line 29: I think this should be returning the newly created record rather than none.

jogilder: And then the code above should be adjusted to allow for that.

laa_crime_application_store_app/routers/v1/application.py
```diff
@@ -102,4 +107,45 @@ async def put_application(
...
+    request: Request,
+    subscriber: Subscriber,
+    db: Session = Depends(get_db),
+):
+    logger.info("CREATING_SUBSCRIBER")
+    existing_subscriber = NotificationService.subscribe(
+        db, subscriber.subscriber_type, subscriber.webhook_url
+    )
+
+    if existing_subscriber is None:
```
jogilder, line 126: It would make sense to have this follow the logic in the 
[application router](https://github.com/ministryofjustice/laa-crime-application-store/blob/d3c89edf53acc29c4f0f7e938d33f95348ad2d5c/laa_crime_application_store_app/routers/v1/application.py#L75) and the [application service](https://github.com/ministryofjustice/laa-crime-application-store/blob/d3c89edf53acc29c4f0f7e938d33f95348ad2d5c/laa_crime_application_store_app/services/v1/application_service.py#L102)

So you only get `None` if there is a duplicate subscriber found.  Although to be honest that logging info and assumption is not great as duplicate application is not the only reason the db commit would fail but that's a completely separate issue.

laa_crime_application_store_app/services/v1/notification_service.py
patrick-laa: Cool, have adjusted

laa_crime_application_store_app/routers/v1/application.py
patrick-laa: Ok, have adjusted.

laa_crime_application_store_app/routers/v1/application.py
```diff
@@ -102,4 +107,45 @@ async def put_application(
...
+        logger.info("SUBSCRIBER_CREATED")
+        return Response(status_code=201)
+    else:
+        logger.info("SUBSCRIBER_ALREADY_EXISTED")
+        return Response(status_code=204)
+
+
+@router.delete("/subscriber")
```
jogilder, line 134: should we have an update as well?

laa_crime_application_store_app/models/subscriber_schema.py
```diff
@@ -0,0 +1,13 @@
...
+
+from sqlalchemy import UUID, Column, String
+
+from laa_crime_application_store_app.data.database import Base
+
+
+class Subscriber(Base):
+    __tablename__ = "subscriber"
+
+    id = Column(UUID, primary_key=True, default=uuid.uuid4)
```
jogilder, line 11: do you think id as primary key or have a composite primary key of subscriber_type and url? You aren't using id in the schema and I understand that as you don't want the subscriber to have to know that.

laa_crime_application_store_app/routers/v1/application.py
jogilder: and maybe put subscriber in its own router so we can do different things with authentication etc if we need to?

laa_crime_application_store_app/services/v1/notification_service.py
```diff
@@ -0,0 +1,64 @@
...
+
+    @staticmethod
+    def unsubscribe(db: Session, subscriber_type: str, webhook_url: str):
+        existing_subscriber = (
+            db.query(Subscriber)
+            .filter(
+                Subscriber.webhook_url == webhook_url,
+                Subscriber.subscriber_type == subscriber_type,
+            )
+            .first()
```
jogilder, line 39: everyone keeps using first - it worries me that we never check to see how many rows filter returns but I guess if we never allow the same combination to be created then we should be ok.

laa_crime_application_store_app/services/v1/notification_service.py
```diff
@@ -0,0 +1,64 @@
...
+        )
+        if existing_subscriber is None:
+            return False
+
+        db.delete(existing_subscriber)
```
jogilder, line 44: wrap in try except

tests/routers/v1/application_test.py
```diff
@@ -132,14 +126,31 @@ def test_post_application_returns_duplicate_error_if_id_already_exists(
     assert response.status_code == 409
 
 
-@patch("laa_crime_application_store_app.internal.notifier.Notifier.notify")
+def test_post_application_notifies_subscribers(
+    client: TestClient, dbsession: Session, seed_subscriber, httpx_mock
+):
+    httpx_mock.add_response(url=seed_subscriber.webhook_url)
```
jogilder, line 132: I am likely being dim but not fully sure I understand how this works.
https://pypi.org/project/pytest-httpx/ says that any response that does not come back will mean the test fails but the examples seem to suggest that you are looking at the response from whatever you do next, ie the client.post. 

Or is it that you get two responses back from this post (our one and the one to the webhook) and you have limited it to checking the webhook one?

> In case more than one response match request, the first one not yet sent (according to the registration order) will be sent.
> 
> In case all matching responses have been sent, the last one (according to the registration order) will be sent.
> 
> You can add criteria so that response will be sent only in case of a more specific matching.
>

jogilder: If that is the case then a comment to explain how the test fails would be good.

jogilder (commented):
Some comments but mainly queries

tests/routers/v1/application_test.py
patrick-laa: So basically there's a hidden assertion as far as I can tell. This line says "If by the end of the test there has been no call made to `seed_subscriber.webhook_url`, throw a wobbly and cause the test to fail." You're right, a comment making this clear makes sense - it wasn't at all clear to me until I deliberately failed the test, but then again _nothing_ in python is clear to me!

laa_crime_application_store_app/services/v1/notification_service.py
patrick-laa: I think your composite primary key suggestion will fix this in this instance, as the database will guarantee there will never be more than one record that matches the criteria.

laa_crime_application_store_app/routers/v1/application.py
patrick-laa: I can't think of a use case for date at the mo. For branch deployments we create on startup and delete on environment deletion, and beyond that I think maybe it's something we add when we have a need for it. But yes, sounds very sensible to have a different router for subscriber.

laa_crime_application_store_app/services/v1/notification_service.py
patrick-laa: Thinking about it, the only way this could fail would be if between retrieving the record on line 40 and deleting it here, the record was deleted by another process. So rather than try/except to fail gracefully in that case, I've added `with_for_update`

jogilder (approved):
LGTM

PR #114, opened 2024-04-10
CRM457-1327: Spike queue

patrick-laa:
Simple task queue:

- Now instead of calling subscriber webhooks directly, the python app saves a QueuedJob object which adds a row to the `que_jobs` table.
- A task queue called Que polls that table, and its workers instantiate a Ruby NotifySubscriberJob with the saved arguments, and execute it, which triggers a request to the webhook
- If the request returns anything other than a 200, an error is raised, causing the job to be retried (with exponential backoff)


patrick-laa:
Closing this as we're re-implementing the whole app in Ruby, not just the queue, and using Sidekiq rather than Que for consistency and better K8s integration.

PR #116, opened 2024-04-12
CRM457-1327: Reimplement in Ruby and switch to Pub/Sub instead of cron

patrick-laa:
- Setup a very simple Rails application that implements the app store logic
- Preserve the python app - run tests and linters on it, just don't push its image to ECR
- Use the existing database - added a migration that will only run if the Python-generated tables don't already exist

What this is missing:
- Role-based permissions


app/models/submission.rb
```diff
@@ -0,0 +1,21 @@
...
+  has_many :submission_versions, dependent: :destroy, foreign_key: "application_id"
+
+  validates :application_state, presence: true
+  validates :application_type, presence: true
+  validates :application_risk, presence: true
+
+  def latest_version
+    # Optimisation: When pulling lists of submissions, this allows a single DB lookup using #includes,
+    # instead of going back to the database for every submission
+    submission_versions.max_by(&:version)
```
dwhenry, line 12: Think can cheat slightly on this one.....

if you order the versions by `version: desc` in the include then you can just call first here and avoid running the ruby `max_by`

so... one option is a default scope on the versions table, but that sucks..

instead:
```
has_many :ordered_submission_versions, -> { order(version: :desc) }, dependent: :destroy, foreign_key: "application_id", table_name: 'SubmissionVersions'
```

and **I think** that will work (untested)

app/models/subscriber.rb
```diff
@@ -0,0 +1,6 @@
+class Subscriber < ApplicationRecord
+  self.table_name = "subscriber"
+
+  validates :webhook_url, presence: true, uniqueness: { scope: :subscriber_type }
+  validates :subscriber_type, presence: true
+end
```
dwhenry, line 6: I still think it might be worth having a failed count stored on this record (i..e how many attempts to send have failed in a row) this means if/when the unsibscribe fails we can use this as a secondary way to unsubscribe (potentially)

app/services/submissions/update_service.rb
```diff
@@ -0,0 +1,32 @@
...
+          submission.current_version += 1 if params[:application]
+          submission.update!(params.permit(:application_state, :application_risk))
+          add_new_version(submission, params) if params[:application]
+        end
+        NotificationService.call(params[:id], role)
+      end
+
+      def add_events(submission, params)
+        params[:events]&.each do |event|
+          next if submission.events.find { _1["id"] == event["id"] }
```
dwhenry, line 17: nit:

prefer:
```
next if submission.events.any? { _1["id"] == event["id"] }
```

db/migrate/20240412111707_allow_long_webhook_types.rb
```diff
@@ -0,0 +1,9 @@
+class AllowLongWebhookTypes < ActiveRecord::Migration[7.0]
+  def up
+    change_column :subscriber, :webhook_url, :string, limit: 250
```
dwhenry, line 3: as a url can be up to 2048 chars (approx) should we just make this a text field.....

helm_deploy/values-uat.yaml
```diff
@@ -23,15 +23,14 @@ securityContext:
...
-   readOnlyRootFilesystem: true
    runAsNonRoot: true
    runAsUser: 10000
    allowPrivilegeEscalation: false
 
 service:
   type: ClusterIP
   port: 80
-  internalPort: 8000
+  internalPort: 3000
```
dwhenry, line 33: would this also need to be changed back to get the python version working? should we chnage the python dockerfile to use 3000 instead so it's not an issue in the future?

dwhenry (approved):
I haven;t looked in detail... too long, but read through the main bits of code and all seems to make sense....

app/models/submission.rb
patrick-laa: Nice! I doubted it would work as `includes` is a big magical, but it does. Thanks!

app/models/subscriber.rb
patrick-laa: Good shout. I've added it, but the auto-delete-above-threshold is only configured to run on dev to avoid screwing us in production.

helm_deploy/values-uat.yaml
patrick-laa: Good idea!

jogilder (commented):
Do the postman tests work?

.circleci/config.yml
```diff
@@ -38,7 +20,8 @@ executors:
           GITHUB_TEAM_NAME_SLUG: laa-assess-a-claim
           REPO_NAME: laa-court-data-api
```
jogilder, line 21: Can we change this as part of this PR?

.circleci/config.yml
```diff
@@ -185,6 +300,12 @@ jobs:
             --build-arg COMMIT_ID=${CIRCLE_SHA1}
             --build-arg APP_BRANCH=${CIRCLE_BRANCH}
 
+          # ========================================================================== #
```
jogilder, line 303: I would be really tempted to take out an the python stuff and just keep it as legacy rather than trying to run both together. We are either going to run with one or the other not both side by side.

jogilder (commented):
I've added in a couple of comments, haven't looked through it. Mostly just want to take out dual purpose for circle and to check port change doesn't cause issues elsewhere, eg postman. Would be good to see what happens with fiddle about /docs and removing that the point in production. 

 Do we get swagger documentation? 

Relying on David to have checked the ruby.

.circleci/config.yml
patrick-laa: Ivan fixed in https://github.com/ministryofjustice/laa-crime-application-store/pull/118 so when that's merged i can rebase.

.circleci/config.yml
patrick-laa: So long as there's python code in the repo I'd prefer to have tests for the python code in CI. However, as soon as we're live with the Ruby code I'd advocate for taking both the python CI and the python code out. At that point it belongs in the git history, not git HEAD.

patrick-laa:
@jogilder 
> Do the postman tests work?

They're passing locally. CI Postman tests are passing but they don't count because they're looking at dev main rather than the branch url - we should ticket up fixing that.

> Do we get swagger documentation?

Not for free sadly, because in Rails you don't declare schemas for parameters in the same way.

.circleci/config.yml
jogilder: Ok, I chatted this through with David and am happy for this to be the way forward.

PR #123, opened 2024-04-16
CRM457-1327: More fixes

patrick-laa:
## Description of change
- Another fix to ping endpoint to match postman expectations
- Handle Socket::ResolutionError the same way as other errors in NotifySubscriber
- Bump UAT and production pod resources to be able to handle a `rails c` alongside ` rails s`
- Allow puma to launch in cluster mode so we can handle a synchronous response to our webhook callbacks in E2E tests
- Use fallback request header to identify client role in non-auth setup.

### DIFF
--- a/app/jobs/notify_subscriber.rb
+++ b/app/jobs/notify_subscriber.rb
@@ -5,15 +5,7 @@ class NotifySubscriber < ApplicationJob
   def perform(subscriber_id, submission_id)
     subscriber = Subscriber.find(subscriber_id)
 
-    response = HTTParty.post(
-      subscriber.webhook_url,
-      headers:,
-      body: {
-        submission_id:,
-      }.to_json,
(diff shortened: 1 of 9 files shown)


ivanELEC (approved):

app/jobs/notify_subscriber.rb
```diff
@@ -5,15 +5,7 @@ class NotifySubscriber < ApplicationJob
...
-      body: {
-        submission_id:,
-      }.to_json,
-    )
-
-    return if response.code == 200
+    return if send_message_to_webhook(subscriber.webhook_url, submission_id)
 
     subscriber.with_lock do
       subscriber.failed_attempts += 1
```
dwhenry, line 11: should we be reseting the failed attempts on a success??? 

to avoid

PR #124, opened 2024-04-18
Disable Production Ingress

ivanELEC:
## Link to relevant ticket

[CRM457-1288](https://dsdmoj.atlassian.net/browse/CRM457-1288)
[CRM457-1289](https://dsdmoj.atlassian.net/browse/CRM457-1289)

## Notes for reviewer
Network Policy and Changes to App Store URLs in Caseworker and Provider App should be complete before merging this

Confirmation of successful call from Production Provider App:
image

Confirmation of successful call from Caseworker App: 

[CRM457-1288]: https://dsdmoj.atlassian.net/browse/CRM457-1288?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CRM457-1289]: https://dsdmoj.atlassian.net/browse/CRM457-1289?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


helm_deploy/values-production.yaml
```diff
@@ -33,24 +33,24 @@ service:
...
+  #   external-dns.alpha.kubernetes.io/set-identifier: laa-crime-application-store-app-laa-crime-application-store-production-green
+  #   external-dns.alpha.kubernetes.io/aws-weight: "100"
+  # hosts:
+  #   - host: laa-crime-application-store-prod.apps.live.cloud-platform.service.justice.gov.uk
+  #     paths:
+  #       - path: /
+  #         pathType: Prefix
+  # tls:
+  #   - hosts:
+  #     - laa-crime-application-store-prod.apps.live.cloud-platform.service.justice.gov.uk
```
dwhenry, line 53: do we want to keep this or can we just delete those lines?

dwhenry: also.. should we also replicate this in UAT as that way we can test any changes that might be related to this (unlikely to be any I know) before relaese

ivanELEC: yeah I was initially going to keep it in just so we had an example of what we had in production if we need to re-enable but I think as long as we have one (Dev) we can get rid

regarding doing this in UAT also, I agree - I was going to do it in another PR but will do it in this one

dwhenry (approved):

PR #125, opened 2024-04-19
CRM457-1146: Add authorization rules

patrick-laa:
## Description of change
- Add a set of rules of what providers and caseworkers can do
- Include what states they can change submissions to and from
- Authorize in a before_action according to the rules

## Link to relevant ticket

[CRM457-1146](https://dsdmoj.atlassian.net/browse/CRM457-1146)

[CRM457-1146]: https://dsdmoj.atlassian.net/browse/CRM457-1146?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/services/authorization_service.rb
+++ b/app/services/authorization_service.rb
@@ -1,0 +1,11 @@
+class AuthorizationService
+  def self.call(subject, category, verb, params, object)
+    rule = Authorization::Rules::PERMISSIONS.dig(subject, category.to_sym, verb.to_sym)
+
+    if rule.respond_to?(:call)
+      rule.call(object, params)
+    else
+      rule
+    end
+  end
+end
--- a/app/controllers/V1/submissions_controller.rb
+++ b/app/controllers/V1/submissions_controller.rb
@@ -15,14 +15,22 @@ module V1
     end
 
     def show
-      render json: Submission.find(params[:id])
+      render json: current_submission
     end
 
     def update
-      Submissions::UpdateService.call(params, current_client_role)
+      Submissions::UpdateService.call(current_submission, params, current_client_role)
       head :created
     rescue ActiveRecord::RecordInvalid
       head :unprocessable_entity
+    end
+
+    def current_submission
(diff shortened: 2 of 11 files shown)


app/services/authorization_service.rb
```diff
@@ -0,0 +1,11 @@
...
+  def self.call(subject, category, verb, params, object)
+    rule = Authorization::Rules::PERMISSIONS.dig(subject, category.to_sym, verb.to_sym)
+
+    if rule.respond_to?(:call)
+      rule.call(object, params)
+    else
+      rule
+    end
+  end
+end
```
dwhenry, line 11: Not sure I love the giant hash....

however the only alternative I have is:
```
class AuthorizationService
  def self.call(subject, category, verb, params, object)
    case [subject, category.to_sym, verb.to_sym]
    when %i[provider subscribers create], %i[provider submissions index], %i[provider submissions show],
        %i[provider submissions create], ...
      true
    when %i[provider subscribers destroy]
      !object || object.subscriber_type == "provider"
    when %i[provider submissions update]
      state_pair_allowed?(object, params, PERMITTED_SUBMISSION_STATE_CHANGES[:provider])
      ...
    end
  end
end
```
which I am not sure is better

dwhenry (approved):
Works... don;t like the giant Hashes

app/services/authorization_service.rb
patrick-laa: Yeah, if it gets much bigger maybe we switch to something like https://github.com/CanCanCommunity/cancancan, but that's probably a bit heavyweight for the current requirement. In the meantime, most of our rules are basically just config at the moment, so I was almost tempted to define it in a yaml file instead of a hash. But again, for the number of rules, the setup to parse and apply yaml is a bit heavy at the mo. I think so long as it's easy to read and understand what each type of user is allowed to do, the hash is the least worst for now.

PR #126, opened 2024-04-19
Amendments from local setup

jsugarman:
## Description of change
Changes following local setup attempt following ruby rewrite

- **Remove actionmailer, mailbox and text railties, routes and controllers**
- **Add sidekiq webui**
- **Add template .env.development file for copying**

## Notes

sidekiq-auth k8s secrets added to dev, UAT and production in readiness


.env.development
```diff
@@ -0,0 +1,21 @@
...
+POSTGRES_PASSWORD=
+POSTGRES_HOSTNAME=localhost
+POSTGRES_NAME=laa_crime_application_store
+
+# To disable need for authentication locally
+# AUTHENTICATION_REQUIRED=false
+
+# Authentication requirements - With authentication enabled
+APP_CLIENT_ID=not-a-real-secret
+CASEWORKER_CLIENT_ID=not-a-real-secret
```
jsugarman, line 15: can remove once patricks PR is in and tested for local setup

.env.development
```diff
@@ -0,0 +1,21 @@
...
+PROVIDER_CLIENT_ID=not-a-real-secret
```
jsugarman, line 16: can remove once patricks PR is in and tested for local setup

.circleci/config.yml
```diff
@@ -7,10 +7,10 @@ orbs:
   aws-ecr: circleci/aws-ecr@9.0.2 # this orb doesn't support OIDC v2, so we use aws-cli to authenticate
 
 executors:
-  python-build-image-executor:
-    resource_class: small
-    docker:
-      - image: cimg/python:3.12.1
+  # python-build-image-executor:
```
patrick-laa, line 10: Rather than comment out the python CI stuff, I think we should delete it. But it's bad to have code in the repo that's not being exercised by CI, so I think if we delete the python CI stuff we should delete the python code from the legacy folder (we were only holding on to it in case the transition to Ruby went horribly wrong)

patrick-laa (approved):

.circleci/config.yml
jsugarman: will do a separate PR after this is merged. just to have something clean to revert if ever needed

PR #127, opened 2024-04-19
Use roles instead of knowing specific IDs

patrick-laa:
## Description of change
Our Entra tokens now contain a "roles" array populated with either `["Provider"]` or `["Caseworker"]`. Using this means we don't need the app store to know the specific client IDs used by provider/caseworker apps, giving us more flexibility.

### DIFF
--- a/spec/requests/authentication_spec.rb
+++ b/spec/requests/authentication_spec.rb
@@ -24,8 +24,10 @@ RSpec.describe "Authentication" do
   context "when an auth token is provided" do
     around do |example|
       ENV["TENANT_ID"] = "TENANT"
+      ENV["APP_CLIENT_ID"] = "APP_STORE"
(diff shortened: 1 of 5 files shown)


spec/requests/authentication_spec.rb
```diff
@@ -74,23 +66,23 @@
       end
 
       context "when caseworker client id is provided" do
-        let(:client_id) { "CASEWORKER" }
+        let(:role) { "Caseworker" }
```
jsugarman, line 69: is there a reason for the role to be initcapped/titleized

patrick-laa: I was a bit surprised - the role "values" are lowercase in Entra, but they come out titlecased in the token. Microsoft works in mysterious ways...

PR #128, opened 2024-04-19
[WIP] Spike redaction code

HettieS:
## Description of change

Copy Crime Apply's approach to redacting data.

This spike PR creates a new table 'redacted_submission_versions' which is linked to 'submission_versions'. When we create the submission version, we also create the redacted record with the same attributes. We then run the application data through a set of rules to replace the specified data with the string "_redacted_".

There is a basic test which validates this behaviour.

Crime Apply's code also has a MetadataWrapper which I have left in the code for now but have commented out as not sure we need it.

## Link to relevant ticket

[CRM457-](https://dsdmoj.atlassian.net/browse/CRM457-1278)


elken:
Closing to clear up k8s resources, can revisit later if desired

PR #130, opened 2024-04-23
App store - add environment to sentry config

jogilder:
## Description of change
Specifiy environment for sentry so it doesn't default to prod

## Link to relevant ticket

[CRM457-](https://dsdmoj.atlassian.net/browse/CRM457-)

### DIFF
--- a/config/initializers/sentry.rb
+++ b/config/initializers/sentry.rb
@@ -1,6 +1,6 @@
 if ENV.fetch("SENTRY_DSN", nil).present?
   Sentry.init do |config|
-    config.environment = ENV.fetch('ENV', 'local')
+    config.environment = ENV.fetch("ENV", "local")
     config.dsn = ENV["SENTRY_DSN"]
     config.breadcrumbs_logger = [:active_support_logger]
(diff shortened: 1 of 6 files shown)


config/initializers/sentry.rb
```diff
@@ -1,5 +1,6 @@
 if ENV.fetch("SENTRY_DSN", nil).present?
   Sentry.init do |config|
+    config.environment = HostEnv.env_name
```
patrick-laa, line 3: HostEnv doesn't exist in this app. You'll need to set an env var in deployment.yaml and deployment-worker.yaml based on .Values.environment, and reference that env var from the ruby code.

jogilder: Using ENV.fetch

patrick-laa:
(deploy-dev is failing because worker deployment config has changed since this PR was opened and the change has been made to an immutable field. You could do a `kubectl delete deployment update-sentry-conf-worker -n laa-crime-application-store-dev` and then re-run the failed CI step to get it to pass. Or you could ignore it and just merge the thing)

PR #132, opened 2024-04-24
Change labels on worker deployment

patrick-laa:
How to verify this fixes stuff:

```
kubectl get pods -n laa-crime-application-store-dev
```

Pick any healthy pod that is neither named `routing-label` nor `laa-crime-application-store-app` and then run

```
kubectl exec -it <pod name> -n laa-crime-application-store-dev -- sh
```

From your shell, if you run the following several times, instead of giving you a 401 every time, it will sometimes give you a "connection refused" error:

```
wget http://laa-crime-application-store-app.laa-crime-application-store-dev.svc.cluster.local
```

This shows that on dev main (with service name `laa-crime-application-store-app`), requests that bypass the ingress are sometimes being routed to worker pods instead of web app pods.

However, if you instead run

```
wget http://routing-label.laa-crime-application-store-dev.svc.cluster.local
```

You should  always receive a 401. That's because it's pointing to this branch, where worker pods have different labels to web app pods, so incoming requests are never routed to them

### DIFF
--- a/helm_deploy/templates/deployment-worker.yaml
+++ b/helm_deploy/templates/deployment-worker.yaml
@@ -10,7 +10,7 @@ spec:
   {{- end }}
   selector:
     matchLabels:
-      {{- include "laa-crime-application-store.selectorLabels" . | nindent 6 }}
+      app:  {{ include "laa-crime-application-store.fullname" . }}-worker
   template:
     metadata:
       {{- with .Values.podAnnotations }}
@@ -18,7 +18,7 @@ spec:
         {{- toYaml . | nindent 8 }}
       {{- end }}
       labels:
-        {{- include "laa-crime-application-store.selectorLabels" . | nindent 8 }}
+        app:  {{ include "laa-crime-application-store.fullname" . }}-worker
     spec:
       serviceAccountName: {{ .Values.service_account.name }}
       {{- with .Values.imagePullSecrets }}


ivanELEC (approved):

ivanELEC:
approved but only just seen @jogilder comment re: checking with Micheal

dwhenry (approved):
I will admit I did like that the app label was appended with `-app` but happy with this

helm_deploy/templates/deployment-worker.yaml
```diff
@@ -10,15 +10,15 @@ spec:
   {{- end }}
   selector:
     matchLabels:
-      {{- include "laa-crime-application-store.selectorLabels" . | nindent 6 }}
+      app:  {{ include "laa-crime-application-store.fullname" . }}-worker
```
jsugarman, line 13: should we remove the `selectorLabels` helper entirely and update `deployment.yaml` to use the same, minus the `-worker`

```yml
  selector:
    matchLabels:
      app: {{ include "laa-crime-application-store.fullname" . }}
  template:
    metadata:
      labels:
        app: {{ include "laa-crime-application-store.fullname" . }}
```

But, this would provide different output of which i am not sure of the impact, namely instead of `app`  the selectorLabels helper interpolates `app.kubernetes.io/name` and adds `app.kubernetes.io/instance` too.

```
app.kubernetes.io/name: {{ include "laa-crime-application-store.name" . }}
app.kubernetes.io/instance: {{ .Release.Name }}

jsugarman (commented):
LGTM but wondering if we should simplify further. see comment

PR #134, opened 2024-04-26
Send payload inside webhook request

patrick-laa:
At the moment, webhooks involve 2 HTTP requests - the app store worker POSTs the submission ID to the client, and the client GETs the submission from the app store to get full details. The idea was that this would allow flexibility. However, since both clients do the GET synchronously, all this is really doing is adding unnecessary complexity and moving parts. If the submission details are included in the body of the initial POST, that means the GET request can be avoided entirely, making the system faster, easier to reason about and less vulnerable to network issues.


jsugarman (approved):
sounds good to me. I guess the subscribers have have changes made to accept the new payload?

PR #136, opened 2024-04-30
Disable modsec in dev

patrick-laa:
False positives break stuff, dev is a low-security environment, and we don't have modsec in UAT or prod anyway due to not having ingresses there.


dwhenry (approved):
can we turn this back on if we use the internal url as they skip mod sec right???

PR #139, opened 2024-05-01
Stop running postman tests

patrick-laa:
The postman tests always fail. This is partially due to the config for authenticating on dev, and partially because the tests haven't been updated to match the new roles logic.

We already have a bunch of request specs here _and_ now E2E tests in the CI pipeline, the postman tests are redundant as we have multiple other things that would break in situations where they would break. So I vote we just remove the postman tests so that we don't get constant CI failure notifications on every merge.


dwhenry (approved):
can we remove the test files as well if we are no longer running

PR #149, opened 2024-05-24
CRM457-1332-Adds deployment for Metabase

imtiazAhmd:
[##](url) Description of change

## Link to relevant ticket

[CRM457-1332](https://dsdmoj.atlassian.net/browse/CRM457-1332)

## Notes for reviewer
Metabase deployment URLs:
There will be just two deployments - one pre-prod in UAT and one prod in prod
~~UAT: laa-crime-forms-metabase-uat.apps.live.cloud-platform.service.justice.gov.uk~~
~~Production: laa-crime-forms-metabase-prod.apps.live.cloud-platform.service.justice.gov.uk~~

To connect a development branch to a metabase UI, [use these instructions](https://dsdmoj.atlassian.net/wiki/spaces/CRM457/pages/4903403612/Metabase+Setup+for+NSCC#Setting-up-the-database-connection-in-the-Metabase-UI)
Currently ONLY UAT Deployment is enabled, when ready we can simply enable in `values-production.yaml`  under `helm_deploy > templates`

Production metabase and UAT metabase RDS created.
RDS dev instance for Metabase will be removed 

[CRM457-1332]: https://dsdmoj.atlassian.net/browse/CRM457-1332?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


helm_deploy/templates/metabase/metabase-deployment.yaml
```diff
@@ -0,0 +1,47 @@
+{{- if .Values.metabase.enabled }}
+{{- $fullName := include "laa-crime-application-store.fullname" . -}}
```
jogilder, line 2: I don't think this works as you haven't got `.fullname` defined anywhere.

helm_deploy/templates/ingress.yaml
```diff
@@ -59,4 +59,16 @@ spec:
               {{- end }}
           {{- end }}
     {{- end }}
+    {{- if .Values.metabase.enabled }}
```
jogilder, line 62: I think this is inside the ingress enabled option - is this what we want? I imagine we need to be able to see the metabase instance even if we have no ingress switched off for the app.

jogilder: I think it might make more sense if we put this bit in it's own ingress file within the metabase templates folder.

jsugarman (commented):
as discussed: 
 - the ingress works fine for metabase system and is an improvement on having it as a second ingress on the app IMO
 - the naming could be simplified to `metabase/deployment.yml` instead of `metabase/metabase-depployment.yml` 
 - we could disable the metabase system on DEV and delete its metabase system RDS instance, instead relying on UAT for all "pre-prod" metabase WebUI work (from which we can point to DEV-main and DEV-branch source databases as needs be)

helm_deploy/templates/metabase/deployment.yaml
```diff
@@ -0,0 +1,47 @@
...
+          requests:
+            cpu: 25m
+            memory: 1Gi
+          limits:
+            cpu: 500m
+            memory: 3Gi
+        securityContext:
+          allowPrivilegeEscalation: false
+          runAsUser: 1000
+          runAsGroup: 1000
```
jsugarman, line 40:
```
securityContext:
  allowPrivilegeEscalation: false
  runAsUser: 1000
  runAsGroup: 1000
```

looking at other config do we need to add this? but not sure what impact it may have
```
  runAsNonRoot: true
  seccompProfile:
    type: RuntimeDefault
  capabilities:
    drop: [ "ALL" ]
```

jogilder: https://dsdmoj.atlassian.net/browse/CRM457-1547

PR #171, opened 2024-06-17
CRM457-1468: Add grafana S3 metrics dashboard

patrick-laa:
## Description of change
Add an S3 metrics dashboard that looks at cloudwatch data for our S3 buckets
Note the bucket IDs are secret, so for the filter (which will be publicly visible) I'm only using the first 3 unique chars of them, leaving 29 characters obscured. This is enough to exclude all other cloud platform buckets for now, but it is possible in future that another one will appear. Either way, to identify which bucket matches which environment we should use:

```bash
 cloud-platform decode-secret -n laa-submit-crime-forms-prod -s s3-bucket-output --skip-version-check
```

The dashboard is visible [here](https://grafana.live.cloud-platform.service.justice.gov.uk/d/laa-crime-store-s3-metrics/laa-non-standard-crime-claims-s3-metrics?orgId=1&var-datasource=P896B4444D3F0DAB8&var-region=default&var-bucket=cloud-platform-5964914c521cc2f2700c40908cdf2e30&var-filterid=laa-submit-crime-forms-uploads&from=now-2d&to=now) - note that until they are used more we won't have much data for production and UAT.


jogilder (approved):
LGTM

PR #184, opened 2024-06-24
[CRM457-1664] allow events only to be updated

dwhenry:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1664)

## Notes for reviewer
When looking at this I initially planned to add a new endpoint, but when looking at the existing code I realised it was already allow (almost) in the current endpoint so chose to modify the rules instead.

Happy to chnage to have a separate endpoint if people feel strongly

### DIFF
--- a/app/services/authorization/rules.rb
+++ b/app/services/authorization/rules.rb
@@ -51,8 +51,7 @@ module Authorization
 
     def self.state_pair_allowed?(object, params, pairs)
       pairs.any? do |pair|
-        object.application_state.in?(pair[:pre]) && params[:application_state].in?(pair[:post]) &&
-          (pair[:criteria].nil? || pair[:criteria].call(params))
+        object.application_state.in?(pair[:pre]) && params[:application_state].in?(pair[:post])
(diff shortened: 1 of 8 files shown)


app/services/authorization/rules.rb
```diff
@@ -48,7 +51,8 @@ module Rules
 
     def self.state_pair_allowed?(object, params, pairs)
       pairs.any? do |pair|
-        object.application_state.in?(pair[:pre]) && params[:application_state].in?(pair[:post])
+        object.application_state.in?(pair[:pre]) && params[:application_state].in?(pair[:post]) &&
+          (pair[:criteria].nil? || pair[:criteria].call(params))
```
patrick-laa, line 55: This is now redundant, I think

patrick-laa (approved):

app/services/authorization/rules.rb
dwhenry: yes.. removed

PR #193, opened 2024-07-02
[CRM457-1630] Rake task to set missing timestamps

dwhenry:
## Description of change
Fix some of the missing data issues (still approx 12 left after these fixes

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1630)

### DIFF
--- a/lib/tasks/missing_version_timestamps.rake
+++ b/lib/tasks/missing_version_timestamps.rake
@@ -1,0 +1,133 @@
+desc "One off tasks to fix missing submission version timestamps"
+namespace :missing_version_timestamps do
+
+  desc "When the latest record missed the required timestamp"
+  task tranch1: :environment do
+    SubmissionVersion.where(created_at: nil)
+
+    puts "==============================="
+    puts "Tranch 1"
+    puts "These are where the submission version is greater than that on any of the events"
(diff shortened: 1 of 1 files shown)


lib/tasks/missing_version_timestamps.rake
```diff
@@ -0,0 +1,133 @@
+desc "One off tasks to fix missing submission version timestamps"
+namespace :missing_version_timestamps do
+
+  desc "When the latest record missed the required timestamp"
+  task tranch1: :environment do
+    SubmissionVersion.where(created_at: nil)
```
patrick-laa, line 6: This is redundant, right? Unless it should be a guard clause?

jsugarman (commented):
need to be wary of using `update` or `save[!]` as these rails methods update timestamps themselves.

I used `update_all`, `update_column` or raw sql to avoid this on my

patrick-laa (approved):
I think it'll work. And if it doesn't, I don't think it can go too far wrong.

lib/tasks/missing_version_timestamps.rake
dwhenry: yes removed

jsugarman (approved):

PR #198, opened 2024-07-03
Add grafana sidekiq dashboard

ivanELEC:
## Description of change

[Basic Sidekiq Metrics in Grafana Dashboard](https://dsdmoj.atlassian.net/browse/CRM457-1486)
[ Example dashboard](https://grafana.live.cloud-platform.service.justice.gov.uk/d/laa-submit-crime-forms-sidekiq-metrics/b5331a96-3a9b-54c0-bbc7-e00f71ef3abe?orgId=1&refresh=5s)


helm_deploy/dashboards/grafana-sidekiq-metrics.yml
```diff
@@ -0,0 +1,910 @@
...
+          "6h",
+          "12h",
+          "24h",
+          "2d",
+          "7d",
+          "30d"
+        ]
+      },
+      "timezone": "browser",
+      "title": "LAA non-standard crime claims / Provider sidekiq metrics",
```
patrick-laa, line 906: Should it just be "Sidekiq metrics"?

helm_deploy/dashboards/grafana-sidekiq-metrics.yml
```diff
@@ -0,0 +1,910 @@
...
+          },
+          "pluginVersion": "10.3.3",
+          "targets": [
+            {
+              "datasource": {
+                "type": "prometheus",
+                "uid": "prometheus"
+              },
+              "editorMode": "code",
+              "expr": "avg(ruby_sidekiq_stats_processed{namespace=\"$namespace\",pod=~\"laa-.+-app-.+\"})",
```
patrick-laa, line 112: This is telling me that in production we've processed nearly 40,000 jobs in the app store. Which I don't think I believe, as we only use jobs for quite specific events that don't happen that often. I wonder if this metric means something different to what we think it means?

helm_deploy/dashboards/grafana-sidekiq-metrics.yml
```diff
@@ -0,0 +1,910 @@
...
+              "expr": "sum(ruby_sidekiq_stats_workers_size{namespace=\"$namespace\"})",
+              "format": "time_series",
+              "intervalFactor": 2,
+              "legendFormat": "Jobs being processed",
+              "refId": "A"
+            }
+          ],
+          "thresholds": [],
+          "timeRegions": [],
+          "title": "Number of jobs actively being processed",
```
patrick-laa, line 810: This is a flat line in Grafana for all time periods for all environments. I don't know that it's telling us what we want it to tell us.

helm_deploy/dashboards/grafana-sidekiq-metrics.yml
ivanELEC: indeed

helm_deploy/dashboards/grafana-sidekiq-metrics.yml
jsugarman: This is "all time" and the high numbers should match what sidekiq webui is displaying for "processed". on assess and submit production envs this should be 66k+ and 63k+. On app store you can use rails console

```
Sidekiq::Stats.new.fetch_stats!
=> {:processed=>39259, :failed=>14, :scheduled_size=>2, :retry_size=>0, :dead_size=>2, :processes_size=>2, :default_queue_latency=>0, :workers_size=>0, :enqueued=>0}
```

patrick-laa: But the only jobs the production app store should have processed is NotifySubscriber jobs, which should be some multiple of the number of SubmissionVersion records, which number in the hundreds. What _are_ the other jobs?

ivanELEC: So SidekiqWorker::Alive has been running since we switched to this API 3 months ago - this is likely the source of the high numbers. But if I'm correct, sidekiq alive is run every 30 seconds (the liveness probe has a period of 30 seconds) if this is the case, the number of jobs executed should be something more in the hundreds of thousands if we assume the worker has been live for most of the 3 months - so it's all a bit mystifying. Honestly though the real fact is that this stat alone doesn't really help us much

helm_deploy/dashboards/grafana-sidekiq-metrics.yml
ivanELEC: Yeah I've changed this to show the stats of jobs processed against failed and restarted. This is the appropriate metric for a chart like this but it seems like the counters are being reset periodically as it keeps dropping to 0 every now and again which a counter shouldn't do (unless Sidekiq::Stats.new.reset) is run

jsugarman (approved):

PR #201, opened 2024-07-04
Crm457 1713 - SPIKE Caseworker: Search and filter - required changes to datastore

ivanELEC:
## Description of change

[SPIKE Caseworker: Search and filter - required changes to datastore](https://dsdmoj.atlassian.net/browse/CRM457-1713)

## Notes for reviewer

- Searches table uses virtual fields to collate search keywords for a given application
- UFN and LAA Reference are special cases as the tsvector functions consider a dash followed by a number a negative number and a slash as an escape so these are searched via the original fields

## Further notes

* we are using weights to split how searching works. 
  - Items with a weight of A require a full match
  - items with a weight of B match the start of string only
* using regexp to remove `-` character from LAA reference to simplify matching
* two know edgecases that don't match when maybe the should:
  - names with a dash followed by a number ie. `Palace Court-1` these will match on `palace`, `court`, `palace-court`, `palace-court-1` but not `1`
  - names with a forward slash i.e. `James/Jim` will match on `James`, `James\Jim` but not `Jim`

### DIFF
--- a/spec/models/search_spec.rb
+++ b/spec/models/search_spec.rb
@@ -157,8 +157,7 @@ RSpec.describe Search do
           expect(search).to eq([prepare.id])
         end
 
-        # TBC: not surrently working due to `/`
-        context 'with additional partial name matches' do
+        context "with additional partial name matches" do
           let(:prepare) do
             [
               build(:submission, :with_pa_version, defendant_name: "Jason/Jim Read").tap(&:save),
@@ -168,12 +167,14 @@ RSpec.describe Search do
             ]
           end
 
-          xit "matches all records on each part of name" do
+          it "matches all records on each part of name" do
+            skip "TBC: not surrently working due to `/`"
+
             expect(search).to eq(prepare[0..2].map(&:id))
           end
         end
 
-        context "search on first part of first name only" do
+        context "when searching on the first part of first name only" do
           let(:query) { "Jason" }
 
           it "returns the record" do
@@ -181,11 +182,12 @@ RSpec.describe Search do
           end
         end
 
-        # TBC: not surrently working due to `/`
-        context "search on last part of first name only" do
+        context "when searching on the last part of first name only" do
           let(:query) { "Jim" }
 
-          xit "returns the record" do
+          it "returns the record" do
+            skip "TBC: not surrently working due to `/`"
+
             expect(search).to eq([prepare.id])
           end
         end
--- a/spec/models/search_spec.rb
+++ b/spec/models/search_spec.rb
@@ -16,7 +16,7 @@ RSpec.describe Search do
       end
     end
 
-    context "when search text is not LAA reference" do
+    context "when search text is LAA reference does not match any records" do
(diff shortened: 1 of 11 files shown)


spec/models/search_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+    end
+
+    context "when search text is 'laa'" do
+      let(:query) { "laa" }
+
+      it "returns the record" do
+        expect(search).to be_empty
+      end
+    end
+
```
dwhenry, line 49: should we add a test for lass reference with name..

i.e. let(:query) { "LAA-123456 joe bloggs" }

@jogilder we have made some assumptions on how matching will work that we likely need to check (or at least document)

spec/models/search_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+      context "with non-standard magistrate application" do
+        let(:prepare) { build(:submission, :with_nsm_version).tap(&:save) }
+
+        it "returns the record" do
+          expect(search).to eq([prepare.id])
+        end
+      end
+    end
+
```
dwhenry, line 69: what do we expect to happen on a partial match??

i.e. 
```
let(:query) { "bob bloggs" }
```

should that match this record?

@jogilder

dwhenry: addition would we be expecting 
```
let(:query) { "job blog" }
```
to match??

spec/models/search_spec.rb
```diff
@@ -0,0 +1,297 @@
...
+          it "returns the record" do
+            expect(search).to eq([prepare.id])
+          end
+        end
+
+        context "when searching on the last part of first name only" do
+          let(:query) { "Jim" }
+
+          it "returns the record" do
+            skip "TBC: not surrently working due to `/`"
```
dwhenry, line 187: one of the remaining issue that I am aware of.

could be solved by replacing `/` with `-` but not 100% sure if this is a good idea.

spec/models/search_spec.rb
```diff
@@ -0,0 +1,297 @@
...
+        it "returns the record" do
+          expect(search).to eq([prepare.id])
+        end
+      end
+
+      context "when searching on the last part of firm name only" do
+        let(:query) { "1984" }
+
+        it "returns the record" do
+          skip "does not match as expects `-1984`"
```
dwhenry, line 290: should this work?

patrick-laa (approved):
Approach looks sensible, test coverage looks good. I don't know the nuts and bolts of exactly how we want queries to work, and I don't know what performance will be like, but this definitely feels like a good place to start from and we can refine further if we hit problems.

spec/models/search_spec.rb
ivanELEC: see what comes out with meeting with Mati - imo probably but would accept it for now (as long as people can search firms against key words, I doubt there will be an instance where people are anything more than mildly inconvenienced from only being able to filter via the keywords and not the number in this edge case)

PR #203, opened 2024-07-08
Add first revision of auto grant view

ivanELEC:
## Description of change

[Provider: Autogrant Analytics Tab](https://dsdmoj.atlassian.net/browse/CRM457-1638)


db/schema.rb
```diff
@@ -22,8 +22,8 @@
     t.datetime "updated_at", precision: nil
     t.jsonb "events"
     t.datetime "created_at", precision: nil
-    t.check_constraint "created_at IS NOT NULL", name: "application_created_at_null", validate: false
-    t.check_constraint "updated_at IS NOT NULL", name: "application_updated_at_null", validate: false
+    t.check_constraint "created_at IS NOT NULL", name: "application_created_at_null"
```
patrick-laa, line 25: This looks wrong - the `validate: false` got added in AddNotNullConstraintsToTimeStamps.

patrick-laa (approved):

db/schema.rb
ivanELEC: rectified by generated a schema.rb file from a fresh with migration
it removes casting user ids to integers which I'm not sure why was in there in the first place

PR #204, opened 2024-07-08
Crm457 1638 autogrant service

ivanELEC:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1638)

Add Translated Service Value and Custom Values to Autogrant View

### DIFF
--- a/db/migrate/20240708153903_create_service_translations.rb
+++ b/db/migrate/20240708153903_create_service_translations.rb
@@ -2,16 +2,16 @@ require 'csv'
 
 class CreateServiceTranslations < ActiveRecord::Migration[7.1]
   def change
-    create_table :service_translations do |t|
+    create_table :translations do |t|
       t.string :key
       t.string :translation
       t.string :translation_type
       t.timestamps
     end
 
-    add_index :service_translations, [:key, :translation_type], unique: true
+    add_index :translations, [:key, :translation_type], unique: true
 
-    ServiceTranslation.upsert_all(translations, unique_by: [:key, :translation_type])
+    Translation.upsert_all(translations, unique_by: [:key, :translation_type])
   end
 
   def translations
--- a/db/migrate/20240708153903_create_service_translations.rb
+++ b/db/migrate/20240708153903_create_service_translations.rb
@@ -5,16 +5,20 @@ class CreateServiceTranslations < ActiveRecord::Migration[7.1]
     create_table :service_translations do |t|
       t.string :key
       t.string :translation
+      t.string :translation_type
       t.timestamps
     end
 
-    add_index :service_translations, :key, unique: true
+    add_index :service_translations, [:key, :translation_type], unique: true
 
-    ServiceTranslation.upsert_all(translations, unique_by: :key)
+    ServiceTranslation.upsert_all(translations, unique_by: [:key, :translation_type])
   end
 
   def translations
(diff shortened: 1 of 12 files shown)


db/migrate/20240708153903_create_service_translations.rb
```diff
@@ -0,0 +1,20 @@
+require 'csv'
+
+class CreateServiceTranslations < ActiveRecord::Migration[7.1]
+  def change
+    create_table :service_translations do |t|
+      t.string :key
+      t.string :translation
+      t.timestamps
```
dwhenry, line 8: if we are going to do this I would have preferred a translations table and then have a translation_type field of service_type

would make it more generic if we need it in the future.

db/schema.rb
```diff
@@ -113,9 +121,14 @@
...
-      (a.application ->> 'service_type'::text) AS service_type
-     FROM (all_events e
+      (a.application ->> 'service_type'::text) AS service_key,
+          CASE
+              WHEN ((a.application ->> 'service_type'::text) = 'custom'::text) THEN ((a.application ->> 'custom_service_name'::text))::character varying
+              ELSE s.translation
+          END AS service
+     FROM ((all_events e
        JOIN application_version a ON (((a.application_id = e.id) AND (a.version = e.submission_version))))
+       JOIN service_translations s ON (((a.application ->> 'service_type'::text) = (s.key)::text)))
```
dwhenry, line 131: this will not show records where a translation doesn't exist....

```
      SELECT e.id,
      e.submission_version,
      e.event_on,
      (a.application ->> 'service_type'::text) AS service_key,
      COALESCE(a.application ->> 'custom_service_name', COALESCE(s.translation, a.application ->> 'service_type'::text)) AS service
     FROM ((all_events e
       JOIN application_version a ON (((a.application_id = e.id) AND (a.version = e.submission_version))))
       LEFT JOIN service_translations s ON (((a.application ->> 'service_type'::text) = (s.key)::text)))
```
> NOTE: This assumes custom_service_name is `NULL` when type isn't custom

db/migrate/20240708153903_create_service_translations.rb
ivanELEC: I think this is a good shout - I could say that we don't really want to keep it like this long term and it would be good to tack it into it's own gem like we discussed about things like this and fee calculations - BUT given we may not get that time I think it's worth doing now

db/schema.rb
ivanELEC: I think the assumption in the note is ok considering those are the soft rules of the provider application 
Thanks for the suggestion!

ivanELEC: updated

db/migrate/20240708153903_create_service_translations.rb
jogilder: @ivanELEC and @dwhenry I'm really reluctant to do this temporarily and I would like to discuss this during the team day. 

I think it would be useful to have an ID for each of these translations as well as the key? We could then use that ID in the json blob and for each client application. 

And in theory the value for the translation should be in a separate table linked via that ID and translation_type should be an id linked to a separate normalised translation type table. 

But can we keep this on hold and look in detail on team day. For now we just use the key

dwhenry (approved):
So the code all appears to work... however it would be goos to see some tests to add confidence the view does what is expected. :)

PR #205, opened 2024-07-08
CMR45-1715: Add search endpoint

jsugarman:
## Description of change
First pass at adding a search and filter endpoint

## Notes for reviewer

The request is expected to:
- make a POST request
- handle a 422 response
- supply the various parameters, including a  zero-indexed `page` (offset) and `per_page` (limit)

**NOTE on parameters:**
 - `not_assigned` and `in_progress` pseudo statuses can be supplied as statuses to filter on
 - client/defendant name will be search for in `defendant` object for PA and `defendants` array for NSM. Only "main" defendant is returned for NSM, in the `data` payload, irrespective of whether name is matched in "main" or additional defendant names.
 - date ranges are beginless and endless when `[submitted|updated]_from` and `[submitted|updated]_to` are not supplied, respectively.
 - `sort_by` and `sort_direction` default to `updated_date: :desc` (i.e. most recently updated first) if not supplied. However `sort_direction` defaults to ascending otherwise so that `A to Z` logic is the default.
 
The response
 - returns `metadata` including the total records plus current page and per_page (reflecting the request only) 
 - returns `data`, which is the view's curated data structure json object
 - returns `raw_data`, which is the standard Submission object as a json object (including latest application version details) 
 
 _We can remove `data` or `raw_data` as we decide, going forward._

We need to wire up the front-end branch to this branch or merge and then do wire up to dev main/UAT etc.


app/controllers/V1/searches_controller.rb
```diff
@@ -0,0 +1,112 @@
...
+                            .gsub("ascending", "asc")
+                            .gsub("descending", "desc")
+    end
+
+    def limit
+      search_params.fetch(:per_page, 20).to_i
+    end
+
+    def offset
+      search_params.fetch(:page, 0).to_i
```
patrick-laa, line 86: As mentioned, you'll need to multiply your offset by `per_page`, so that for the first page of results the offset is 0, for the second it's 20, for the third it's 40 and so on.

db/views/searches_v02.sql
```diff
@@ -0,0 +1,15 @@
+SELECT
+  app.id,
+  app_ver.id as application_version_id,
+  app_ver.application ->> 'laa_reference' as laa_reference,
+  app_ver.application -> 'firm_office' ->> 'name' as firm_name,
+  (app_ver.application -> 'defendant' ->> 'first_name') || ' ' || (app_ver.application -> 'defendant' ->> 'last_name') as client,
+  app_ver.search_fields,
+  app.has_been_assigned_to,
+  app.created_at as date_submitted,
+  app.updated_at as date_updated, -- events can happen in caseworker after this? should we be checking max event time as well???
```
patrick-laa, line 10: Events aren't a separate table, so every time an event is added to the event array `updated_at` will be bumped, but it won't be the timestamp of the last update, it will be the timestamp of the last time an event completed syncing via a background job. So ideally yeah, you'd use max event time OR app_ver.created_at, whichever is later

spec/requests/authorization_spec.rb
```diff
@@ -1,5 +1,42 @@
 require "rails_helper"
 
+RSpec.shared_examples "subscriber and unsubscriber" do
```
patrick-laa, line 3: Isn't this already covered in the *_subscriber_spec.rb specs? It doesn't seem like much of this belongs in an authorization spec.

jsugarman: so I found that when i wanted to test that search can only be used caseworker that none of the other authorization rules were explcitly being tested, so i went off on a tangent of excercising the subsccriber authorization rules from an integration spec point of view as a starting point. I was going to add for submissions as well before i pulled myself out of this tangent, and just tested the search rule i had added.

happy to remove but it feels like we do not actually integration test the authorization rules, despite coverage being happy. What say you?

app/controllers/V1/searches_controller.rb
jsugarman: As mentioned in discussions sicne, the offset/page and limit/per_page are supplied by the client that is then responsible for telling the API what it wants - so i don't think this logic applies?!

db/views/searches_v02.sql
jsugarman: Will look into this a bit more. Means more gnarly SQL i am guessing.

app/controllers/V1/searches_controller.rb
```diff
@@ -0,0 +1,112 @@
...
+    end
+
+    def sort_clause
+      return { date_updated: :desc } unless search_params[:sort_by]
+
+      { sort_by => sort_direction }
+    end
+
+    def sort_by
+      search_params.fetch(:sort_by, :date_updated)
```
patrick-laa, line 76: Caseworker is sending `last_updated` rather than `date_updated`

app/controllers/V1/searches_controller.rb
```diff
@@ -0,0 +1,112 @@
...
+        :per_page,
+        :sort_by,
+        :sort_direction,
+        :submission_type,
+        :submitted_from,
+        :submitted_to,
+        :updated_from,
+        :updated_to,
+        :caseworker_id,
+        :status,
```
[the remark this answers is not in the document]
jsugarman: was of two minds about this but happy to change to status_with_assignment to semantically note the the difference

app/controllers/V1/searches_controller.rb
```diff
@@ -0,0 +1,112 @@
...
+    end
+
+    def search_params
+      params.permit(
+        :query,
+        :page,
+        :per_page,
+        :sort_by,
+        :sort_direction,
+        :submission_type,
```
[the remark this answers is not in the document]
jsugarman: there is a fair bit of naming issue relating to submission versus application and application versus claim. In the context of app store that is even more confused due to the underlying table names and model names, not to mention controller names (ApplicationController). I prefer submission type to avoid the ambiguity, at least in the app store, but ideally in the caseworker too.

spec/requests/authorization_spec.rb
patrick-laa: Fair play, happy to leave this in.

app/controllers/V1/searches_controller.rb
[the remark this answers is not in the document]
patrick-laa: So the reason I went with `application_type` was that that's what the field is in the app store database, the caseworker database and in all the payloads that currently go back and forth between app store and provider/caseworker. Calling it `submission_type` in this one instance feels odd, even though I agree it's a better name overall. But I don't want to bikeshed so I'll change on my end.

app/controllers/V1/searches_controller.rb
jsugarman: right, yes, got what you mean now. i will use page and per_page, 1 indexed, and calculate the offset.

jsugarman: applied changes now

config/routes.rb
```diff
@@ -23,6 +23,8 @@
     resources :subscribers, only: %i[create]
     delete :subscribers, to: "subscribers#destroy"
 
+    resource :search, only: %[create]
```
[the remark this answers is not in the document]
jsugarman: implemented (on other branch) `v1/search` --> `v1/submissions/searches`

app/controllers/V1/searches_controller.rb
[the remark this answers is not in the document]
jsugarman: its ok, i've changed my end now

app/controllers/V1/searches_controller.rb
jsugarman: so we currently have `date_submitted` and `date_updated` which i quite like the symmetry of and i have found the different naming a useful distinction. This field needs more work anyway (re what date should actually be presented here) which would like to do post this PR in any event. Can we revisit once this is in?

db/views/searches_v02.sql
jsugarman: I have analysed the data on production and found that of 219 submissions:

#### number of submissions with greater than 60 sec diff between event and updated_at
- 17 over a minute difference between app updated and latest event time
- [only] 3 if excluding those updated on 1 July 2024 by a backfill operation
- app_updated_at is always later than latest event if there is a diff.

#### number of submissions with greater than 60 sec diff between event and app_ver.created
- 23 over a minute difference
- 20 excluding those updated on 1 July 2024 by a backfill operation
- [only] 4 where app_ver_created is actually present  i.e. need to backfill app_ver timestamps

PR #206, opened 2024-07-09
CRM457-1726: Do not error out if caseworker submits late and redundant event payload

patrick-laa:
## Description of change
If request would not actually modify state, short-circuit with 204

Also make services a bit more modular so it's clearer what logic is running in what circumstances

This is an alternative approach to solve the same problem as https://github.com/ministryofjustice/laa-assess-crime-forms/pull/601. Basically if the caseworker app is late to the party trying to tell the app store about an event it already knows about, we don't want errors raised.

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1726)

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -12,7 +12,7 @@ class ApplicationController < ActionController::API
   end
 
   def authorize!
-    allowed = AuthorizationService.call(
+    auth_result = AuthorizationService.call(
       @current_client_role,
       controller_name,
       action_name,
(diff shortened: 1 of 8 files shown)


app/controllers/application_controller.rb
```diff
@@ -12,15 +12,20 @@ def authenticate!
...
       params,
       authorization_object,
     )
 
-    head(:forbidden) unless allowed
+    case auth_result
+    when :forbidden
+      head(:forbidden)
+    when :no_op
+      head(:no_content)
```
patrick-laa, line 27: I chose a 204 No Content over a 304 Not Modified because 304 is very commonly used and, despite the more appropriate-seeming title, has its own quite specific meaning that is supposed only to apply to HEAD and GET requests, so overriding its meaning felt like it could get confusing.

patrick-laa:
Closing as handled by caseworker app

PR #208, opened 2024-07-09
CRM457-1555: Restore files that were deleted in previous commit

elken:
Restores files previously removed in https://github.com/ministryofjustice/laa-crime-application-store/commit/1e963e18f95484e631dad837fe41698ab42265e1.

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1555)

### DIFF
--- a/db/views/version_events_v01.sql
+++ b/db/views/version_events_v01.sql
@@ -1,0 +1,7 @@
+SELECT
+  id,
+  application_type,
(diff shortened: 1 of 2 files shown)


patrick-laa (approved):
This is a possible downside of the `scenic` gem which we're using to handle view definitions - they're like migrations rather than schema.rb, in that you need to keep every version of every view you ever had available in the source code, even if it's since been deleted, in order to prevent migrations from breaking, and that's not totally obvious.

PR #209, opened 2024-07-09
[CRM457-1733] update dev env config template

robinsk1:
## Description of change

[[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-XXX)](https://dsdmoj.atlassian.net/browse/CRM457-1733)

### Before changes:
`env.development.local` created without default redis_url

### After changes:
`env.development.local` includes default redis_url

provider, app-store, caseworker should be able to fetch/update CRMs between them in a new dev env setup

### DIFF
--- a/.env.development
+++ b/.env.development
@@ -9,6 +9,7 @@ POSTGRES_NAME=laa_crime_application_store
 
 # To disable need for authentication locally
 # AUTHENTICATION_REQUIRED=false
+REDIS_URL=redis://localhost:6379/1
 
 # Authentication requirements when authentication enabled
 APP_CLIENT_ID=not-a-real-secret


.env.development
```diff
@@ -9,6 +9,7 @@ POSTGRES_NAME=laa_crime_application_store
 
 # To disable need for authentication locally
 # AUTHENTICATION_REQUIRED=false
+REDIS_URL=redis://localhost:6379/1
```
jsugarman, line 12: local setup has the submit application using `redis://localhost:6379/1` [as here](https://github.com/ministryofjustice/laa-submit-crime-forms/blob/main/config/initializers/sidekiq.rb#L31) which i think could cause "job confusion" for sidekiq if we use it here. I think, in any event, it gets overridden in the sidekiq [initializer here](https://github.com/ministryofjustice/laa-crime-application-store/blob/main/config/initializers/sidekiq.rb#L19-L26).

robinsk1: cheers - will have another look at this

PR #216, opened 2024-07-12
CMR45-1715: Add search endpoint

jsugarman:
## Description of change
First pass at adding a search and filter endpoint

## Notes for reviewer

The request is expected to:
- make a POST request
- handle a 422 response
- supply the various parameters, including a  ~~zero~~one-indexed `page` (offset) and `per_page` (limit)

**NOTE on parameters:**
 - `not_assigned` and `in_progress` pseudo statuses can be supplied as statuses to filter on
 - client/defendant name will be search for in `defendant` object for PA and `defendants` array for NSM. Only "main" defendant is returned for NSM, in the `data` payload, irrespective of whether name is matched in "main" or additional defendant names.
 - date ranges are beginless and endless when `[submitted|updated]_from` and `[submitted|updated]_to` are not supplied, respectively.
 - `sort_by` and `sort_direction` default to `updated_date: :desc` (i.e. most recently updated first) if not supplied. However `sort_direction` defaults to ascending otherwise so that `A to Z` logic is the default.
 
The response
 - returns `metadata` including the total records plus current page and per_page (reflecting the request only) 
 - returns `data`, which is the view's curated data structure json object
 - returns `raw_data`, which is the standard Submission object as a json object (including latest application version details) 
 
 _We can remove `data` or `raw_data` as we decide, going forward._

We need to wire up the front-end branch to this branch or merge and then do wire up to dev main/UAT etc.


jsugarman:
see unresolved comments on https://github.com/ministryofjustice/laa-crime-application-store/pull/205

app/controllers/V1/submissions/searches_controller.rb
```diff
@@ -0,0 +1,137 @@
+module V1
+  module Submissions
+    class SearchesController < ApplicationController
+      def create
+        @data = search_query
```
robinsk1, line 5: another option might be to extract theese private methods out to their own concern - maybe as a service class sending through `search_params` as an arg?

robinsk1 (approved):

patrick-laa (approved):
I'm going to put my UI branch (which points at this branch) into QA to get some feedback from Mary, cos I think this is really solid now).

app/controllers/V1/submissions/searches_controller.rb
```diff
@@ -0,0 +1,144 @@
...
+      end
+
+      def sort_clause
+        return "date_updated desc" unless search_params[:sort_by]
+        raise "Unsortable column \"#{sort_by}\" supplied as sort_by argument" unless SORTABLE_COLUMNS.include?(sort_by.downcase)
+
+        if sort_by == "date_updated"
+          "#{sort_by} #{sort_direction}"
+        else
+          "LOWER(#{sort_by}) #{sort_direction}"
```
dwhenry, line 91: so I don;t think this `lower` shouldn't be needed... 

as the most fields are stored in lower case and those that aren't (search_field) are already converted to lowercase in the virtual column definition

dwhenry (approved):
definitely a step in the right direction.

Lots to iterate on but can't see any reasons to not merge

app/controllers/V1/submissions/searches_controller.rb
jsugarman: This is specifically for sorting the found/returned results. I certainly find that locally a lower case name appears after or before ALL upper case names, for example. The database collation type of which ever database is in use could also be the issue but a simple "lower" avoids the problem across any database.

app/controllers/V1/submissions/searches_controller.rb
jsugarman: I was thinking of doing this if it gets much more complex yes, but have left it as "shameless green" for now.

db/views/searches_v02.sql
```diff
@@ -0,0 +1,53 @@
...
+SELECT
+  app.id,
+  app_ver.id as application_version_id,
+  app_ver.application ->> 'laa_reference' as laa_reference,
+  app_ver.application -> 'firm_office' ->> 'name' as firm_name,
+  def.client_name,
+  app_ver.search_fields,
+  app.has_been_assigned_to,
+  app.created_at as date_submitted,
+  app.updated_at as date_updated, -- events can happen in caseworker after this? should we be checking max event time as well???
```
jsugarman, line 40: use last_updated_at here once other pr merged.

PR #218, opened 2024-07-15
CRM457-1715: Add last_updated_at column, backfill and populate on submission create or update

jsugarman:
## Description of change
Add last_updated_at column, backfill and populate on submission create or update

To allow searching and filtering on a top level
date value.

### DIFF
--- a/app/services/submissions/creation_service.rb
+++ b/app/services/submissions/creation_service.rb
@@ -13,7 +13,8 @@ module Submissions
             version: 1,
           )
 
-          submission.update_columns(last_updated_at: submission.created_at)
+          last_updated_at = params.dig(:application, :updated_at)&.to_time || submission.created_at
+          submission.update_columns(last_updated_at:)
         end
         NotificationService.call(params[:application_id], role)
       end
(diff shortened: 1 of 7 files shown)


app/services/submissions/update_service.rb
```diff
@@ -21,6 +21,7 @@ def add_events(submission, params, save: false)
           event["updated_at"] ||= event["created_at"]
 
           submission.events << event.as_json
+          submission.last_updated_at = event["created_at"]
```
patrick-laa, line 24: I think we probably want safeguards that when an array of events is passed in, we use the maximum `created_at`. There doesn't seem to be anything explicit in the payload construction on the caseworker end that the most recent event will be at the end of the array.

app/services/submissions/creation_service.rb
```diff
@@ -12,6 +12,8 @@ def call(params, role)
             application: params[:application],
             version: 1,
           )
+
+          submission.update_columns(last_updated_at: submission.created_at)
```
dwhenry, line 16: ideally we should pull this from `params[:application]['updated_at']` but that is currenlty only populated for NSM applications... I would push to add it to the PA payload to include it before merging this.

This guards against delays in sidekiq processing affecting the recorded order of submission.

patrick-laa (approved):

PR #220, opened 2024-07-16
CRM457-1752: Backfill crm4 application jsob blobs to have timestamps

jsugarman:
## Description of change
Backfill crm4 application jsob blobs to have timestamps

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1752)

Update CRM4 application blobs to have the timestamps if they do not

It has been agreed to update these timestamps to be the created_at for
the application record itself because, in ~95% of cases, this will be
within seconds of the actual timestamp of submission from provider.

We set the json's updated_at to reflect this "almost submission" timestamp
We set the json's created_at to the same, despite this being inaccurate, as it
 - makes it easier to identify in the future if necessary
 - would be onerous to extract actual created_at dates from provider app and serve
   little purpose.

Note that these timestamps are now being set on submission from provider, going forward,
in any event, so we only need to update those where they are null.


ivanELEC:
Just want to double check - we won't have to consider doing this out of hours or anything to prevent conflicts because in normal functionality we dont update SubmissionVersions, we create new ones right? The only function happening on SubmissionVersions is getting them once they are created

jsugarman:
> Just want to double check - we won't have to consider doing this out of hours or anything to prevent conflicts because in normal functionality we dont update SubmissionVersions, we create new ones right? The only function happening on SubmissionVersions is getting them once they are created

I do not believe out of hours is needed, no. This simply updates existing all submisionversions to have the timestamps they are now getting going forward from a PR merged into provider yesterday. This is needed to ensure i can then backfill a new field last_updated_at (other open PR atm), using either latest event or latest submission_version.application.created_at|updated_at

db/migrate/20240716095916_backfill_application_version_application_json_blob_timestamps.rb
```diff
@@ -0,0 +1,32 @@
...
+    SubmissionVersion
+      .joins(:submission)
+      .where(submission: { application_type: 'crm4' })
+      .where("application->>'created_at' IS NULL AND application->>'updated_at' IS NULL")
+      .each do |ver|
+        ver.application['created_at'] = ver.submission.created_at
+        ver.application['updated_at'] = ver.submission.created_at
+        ver.save(touch: false)
+
+        sleep(0.001)  # throttle
```
robinsk1, line 29: just curious but why do we invoke that throttle (think i've seen it in other PRs as well) - is that a k8s related thing?

jsugarman: @robinsk1 the idea comes from here https://github.com/ankane/strong_migrations?tab=readme-ov-file#backfilling-data and is just my habit for backfilling, and other migration tasks. we could implement the gem itself to warn of unsafe migrations but it can be a bit onerous

PR #233, opened 2024-07-30
Crm457 1547 security context cleanup

ivanELEC:
## Description of change

[Check securityContext for all services](https://dsdmoj.atlassian.net/browse/CRM457-1547)

## Notes for reviewer

- Put Security Context default settings under a re-usable helper template
- Add information about the settings

### DIFF
--- a/helm_deploy/templates/_helpers.tpl
+++ b/helm_deploy/templates/_helpers.tpl
@@ -56,6 +56,18 @@ app: laa-crime-application-store-prometheus
 {{- end }}
 
 {{/*
+Default Security Context
+*/}}
+{{- define "helm_deploy.defaultSecurityContext" -}}
+runAsNonRoot: true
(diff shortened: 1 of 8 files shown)


helm_deploy/templates/_helpers.tpl
```diff
@@ -55,6 +55,18 @@ This label allows the allow-prometheus-scraping NetworkPolicy to identify pods t
 app: laa-crime-application-store-prometheus
 {{- end }}
 
+{{/*
+Default Security Context
+*/}}
+{{- define "helm_deploy.defaultSecurityContext" -}}
```
jsugarman, line 61: for consistency it should be prefixed with the what other "methods" are prefixed with, namely 

```
{{- define "laa-crime-application-store.defaultSecurityContext" -}}
```

jsugarman (approved):

PR #240, opened 2024-08-02
Crm457-1827/fix expired

dwhenry:
## Description of change
Yet another data fix... fun right..

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1827)


lib/tasks/adjust.rake
```diff
@@ -71,4 +71,23 @@ namespace :adjust do
...
+  desc "Fix updated at on last sent back before expiry as expiry version does not update"
+  task fix_expired: :environment do
+    Submission.where(application_state: 'expired').each do |submission|
+      version = submission.ordered_submission_versions[1]
+      wrong_event = submission.events.filter { _1['submission_version'] == (version.version - 1) }
+                              .sort_by { _1['updated_at'] }.last
+      right_event = submission.events.filter { _1['submission_version'] == (version.version - 1) && _1['event_type'] != 'expiry' }
+                              .sort_by { _1['updated_at'] }.last
+
+      if wrong_event['update_at'] != right_event['updated_at'] && version.application['updated_at'] == wrong_event['updated_at']
```
jsugarman, line 84:
```suggestion
      if wrong_event['updated_at'] != right_event['updated_at'] && version.application['updated_at'] == wrong_event['updated_at']
```

lib/tasks/adjust.rake
```diff
@@ -71,4 +71,23 @@ namespace :adjust do
...
+        puts "Fixing incorrect time #{submission.id}:#{version.id} from #{wrong_event['update_at']} to #{right_event['updated_at']}"
```
jsugarman, line 85:
```suggestion
        puts "Fixing incorrect time #{submission.id}:#{version.id} from #{wrong_event['updated_at']} to #{right_event['updated_at']}"
```

spec/lib/tasks/adjust_fix_expired_spec.rb
```diff
@@ -0,0 +1,40 @@
+require "rails_helper"
+
+RSpec.describe "adjust:fix_expired" do
+  before do
+    load "spec/lib/tasks/adjust_fix_expired_data.rb"
+  end
+
+  let(:klass) do
+    Class.new(ApplicationRecord) do
+      self.table_name = :processing_times
```
jsugarman, line 10: why are we using this view for the test?

dwhenry: Because it is the reason I was able to find this bug (The processing time was incorrectly coming back as 0 so it felt right to use in the the check as it proves the issue is fixed.

jsugarman (approved):

PR #244, opened 2024-08-07
Add lock to add_events method

ivanELEC:
## Description of change

[Status not being updated in JSON blob for expired/autogranted actions from CW](https://dsdmoj.atlassian.net/browse/CRM457-1827)

## Notes for reviewer
UpdateService#add_events did not lock the submission and therefore when executed alone could cause issues with the order of the submission being updated (causing data loss) - this ticket ensures the submission is locked add_events is occuring to ensure updates happen in the correct order

### DIFF
--- a/app/services/submissions/update_service.rb
+++ b/app/services/submissions/update_service.rb
@@ -12,20 +12,22 @@ module Submissions
       end
 
       def add_events(submission, params, save: false)
-        submission.events ||= []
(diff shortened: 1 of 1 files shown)


ivanELEC:
Closing because handled in this PR: https://github.com/ministryofjustice/laa-crime-application-store/pull/245/

PR #257, opened 2024-08-21
CRM457-1905: Check submission state

patrick-laa:
## Description of change
We found an issue where submissions were sent to the app store with status 'draft'. But we didn't spot it immediately because we were silently 'correcting' the status in the state field, but not in the "blob".  Let's be more picky, and fail out immediately if we spot impermissible data

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1905)

### DIFF
--- a/app/services/authorization/rules.rb
+++ b/app/services/authorization/rules.rb
@@ -9,7 +9,7 @@ module Authorization
         submissions: {
           index: true,
           show: true,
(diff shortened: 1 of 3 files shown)


elken (approved):

elken:
Worth adding a case for the `draft` state that we hit?

PR #260, opened 2024-08-22
CRM457-1803: Make state consistent

patrick-laa:
## Description of change
Rename state column for consistency
Remove 'provider_requested' status

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1803)

### DIFF
--- a/app/models/submission.rb
+++ b/app/models/submission.rb
@@ -5,7 +5,7 @@ class Submission < ApplicationRecord
            foreign_key: "application_id",
            class_name: "SubmissionVersion"
 
-  validates :application_state, presence: true
+  validates :state, presence: true
   validates :application_type, presence: true
   validates :application_risk, presence: true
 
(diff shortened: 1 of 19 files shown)


app/models/submission.rb
```diff
@@ -14,7 +14,8 @@ def latest_version
   end
 
   def as_json(*)
-    super(only: %i[application_state application_risk application_type updated_at created_at]).merge(
+    super(only: %i[application_risk application_type updated_at created_at]).merge(
+      application_state: state,
```
elken, line 18: Should this not be `state`?

app/services/submissions/update_service.rb
```diff
@@ -5,7 +5,7 @@ def call(submission, params, role)
         submission.with_lock do
           submission.current_version += 1
           EventAdditionService.call(submission, params)
-          submission.update!(params.permit(:application_state, :application_risk))
+          submission.update!(params.permit(:application_risk).merge(state: params[:application_state]))
```
elken, line 8: Should this not be `state` too?

app/models/submission.rb
patrick-laa: Ah, no, because that would be a breaking change to the API, which is out of scope for this ticket. The JSON payloads passed around between the apps still use a field called `application_state`, and will do until we make a version 2 of the API. All we're doing it handling how that field is represented internally within the apps.

app/services/submissions/update_service.rb
patrick-laa: Similar thing - the HTTP API hasn't changed, so the param name will still be `application_state`

PR #295, opened 2024-10-15
CRM457-2094: Make job enqueueings atomic

patrick-laa:
## Description of change
So that if there's a Redis issue we won't end up in an inconsistent state.
Note a few refactorings to make sure things stay locked in the right places etc

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2094)

### DIFF
--- a/app/jobs/notify_subscriber.rb
+++ b/app/jobs/notify_subscriber.rb
@@ -26,9 +26,7 @@ class NotifySubscriber < ApplicationJob
     subscriber.failed_attempts += 1
     subscriber.save!
 
-    return false if delete_subscriber(subscriber)
-
-    true
+    !delete_subscriber(subscriber)
   end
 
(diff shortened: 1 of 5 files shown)


app/jobs/notify_subscriber.rb
```diff
@@ -1,29 +1,35 @@
...
-    subscriber.with_lock do
-      subscriber.failed_attempts += 1
-      subscriber.save!
-    end
+  def handle_failure(subscriber)
+    subscriber.failed_attempts += 1
+    subscriber.save!
 
-    return if delete_subscriber(subscriber)
+    return false if delete_subscriber(subscriber)
```
elken, line 29: `return delete_subscriber(subscriber)` since it returns a bool?

elken (approved):

PR #296, opened 2024-10-15
Bump metabase version

patrick-laa:
Because snyk is complaining about a vulnerability in the current one.


elken (approved):
I'll raise this upstream tomorrow

PR #302, opened 2024-10-18
Raise more errors and clarify what is going on

patrick-laa:
## Description of change
There are some errors we haven't been catching but should
And the logic was a little hard to follow

### DIFF
--- a/app/jobs/notify_subscriber.rb
+++ b/app/jobs/notify_subscriber.rb
@@ -13,17 +13,24 @@ class NotifySubscriber < ApplicationJob
   def perform(subscriber_id, submission)
     raise_error = false
     subscriber = Subscriber.find(subscriber_id)
-    raise_error = handle_failure(subscriber) unless send_message_to_webhook(subscriber.webhook_url, submission)
+    unless send_message_to_webhook(subscriber.webhook_url, submission)
+      raise_error = handle_failure_and_decide_whether_to_raise_error(subscriber)
+    end
 
     submission.update!(notify_subscriber_completed: true)
     raise ClientResponseError, "Failed to notify subscriber about #{submission.id} - #{subscriber.webhook_url} returned error" if raise_error
   end
 
-  def handle_failure(subscriber)
+  def handle_failure_and_decide_whether_to_raise_error(subscriber)
     subscriber.failed_attempts += 1
     subscriber.save!
 
-    !delete_subscriber(subscriber)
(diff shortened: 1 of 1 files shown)


app/jobs/notify_subscriber.rb
```diff
@@ -13,17 +13,24 @@ def self.perform_later(subscriber_id, submission)
   def perform(subscriber_id, submission)
     raise_error = false
     subscriber = Subscriber.find(subscriber_id)
-    raise_error = handle_failure(subscriber) unless send_message_to_webhook(subscriber.webhook_url, submission)
+    unless send_message_to_webhook(subscriber.webhook_url, submission)
+      raise_error = handle_failure_and_decide_whether_to_raise_error(subscriber)
```
elken, line 17: `handle_failure_maybe_raise`?

app/jobs/notify_subscriber.rb
```diff
@@ -37,17 +44,20 @@ def send_message_to_webhook(webhook_url, submission)
...
   end
 
+  # Returns a boolean indicating whether the subscriber has been deleted.
   def delete_subscriber(subscriber)
     deletion_threshold = ENV.fetch("SUBSCRIBER_FAILED_ATTEMPT_DELETION_THRESHOLD", nil)
 
-    return unless deletion_threshold&.to_i&.positive? && deletion_threshold.to_i <= subscriber.failed_attempts
+    return false unless deletion_threshold&.to_i&.positive? && deletion_threshold.to_i <= subscriber.failed_attempts
 
     subscriber.destroy!
```
elken, line 59: Why doesn't `destroy!` return said bool?

elken (approved):

app/jobs/notify_subscriber.rb
patrick-laa: Ah, because destroy either returns a `Subscriber` object or raises an exception. Obviously a `Subscriber` is truthy, but for the sake of clarity I'm trying to use booleans `true/false` rather than `object/nil`

app/jobs/notify_subscriber.rb
patrick-laa: Ooh that's nicer.

elken: Naming things is hard, I do kinda wish Ruby had some kind of standard documentation thing like JSDoc

PR #311, opened 2024-10-30
CRM457-2176: Update definition of submission_by_date view

ivanELEC:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2176)

Needed to fix NSM Submissions Dashboard


elken (approved):

db/views/submissions_by_date_v02.sql
```diff
@@ -0,0 +1,16 @@
+SELECT
```
patrick-laa, line 1: Think we might be missing the migration that actually applies this view?

ivanELEC: yeah looking into this now - scenic being weird its not creating the versioned files properly on its own

ivanELEC: potentially the view was renamed and that's messing up scenic working as expected - I can just manually create the correct migrations so going to do that instead

PR #312, opened 2024-10-30
CRM457-2214: Sanitize the search query first

elken:
## Description of change
An issue was raised in which PG doesn't like having unmatched brackets and other statements in the search query, so we attempt to remove them first.

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2214)

### DIFF
--- a/app/models/search.rb
+++ b/app/models/search.rb
@@ -10,17 +10,34 @@ class Search < ApplicationRecord
   def self.where_terms(string)
     return all unless string
 
+    # Split each string into words by one-or-more whitespace characters
     sub_strings = string.strip.downcase.split(/\s+/).map do |str|
-      if str.start_with?("laa-")
-        clean_str = str.sub(/\Alaa-/, "laa")
-        "#{clean_str}:A"
-      elsif /\A\d+\/\d+\z/.match?(str) then "#{str}:A"
-      elsif /\A\d+{6}\z/.match?(str) then "#{str}:*"
+      # We escape with single quotes
+      str = str.gsub("'", "''")
+
+      # Is this an LAA reference? (LAA-123ABC)
+      if str.start_with?("laa-") then %('''#{str.gsub(/\Alaa-/, 'laa')}''':A)
+
+      # Is this a UFN? (311223/001)
+      elsif /\A\d+\/\d+\z/.match?(str) then %('''#{str}''':A)
+
+      # Is this MAYBE a UFN? (6 digits for the first part, 9 digits for the full)
+      elsif /\A\d{6}\z/.match?(str) then "#{str}:*"
+
+      # Don't know what it is, but escape it anyway.
+      # We've already checked for UFN and LAA reference, so we give it weight B
       else
-        "#{str.tr('/', '-')}:*B"
+        %('''#{str.tr('/', '-')}''':*B)
       end
     end
 
+    # The way the lexemes are setup in the search vector means a typical record ends up looking like
+    #
+    # '-123456':8A '010124/001':6A 'laa-123456':10A 'jason':2B 'jason-jim':1B 'jim':3B  'read':4B,5B
+    #
+    # In which we have the expected fields converted to lexemes, however in the case of the "Jason/Jim"
+    # first name; other lexemes are generated for both the full and partial matches of both names because
+    # we convert the / to a - which gets treated as a word boundary.
     where("searches.search_fields @@ to_tsquery('simple', ?)", sub_strings.compact.join(" & "))
   end
 end
--- a/spec/requests/submissions_search_spec.rb
+++ b/spec/requests/submissions_search_spec.rb
@@ -640,5 +640,160 @@ RSpec.describe "Submission search" do
         expect(laa_references).to match(%w[LAA-CCCCCC LAA-BBBBBB LAA-AAAAAA])
       end
     end
+
+    context "when searching for queries that may be invalid" do
+      before do
+        create(:submission,
+               :with_pa_version,
+               laa_reference: "LAA-AAAAAA",
+               defendant_name: "Fred Arbor",
+               firm_name: "Aardvark & Co")
+
+        create(:submission,
+               :with_pa_version,
+               laa_reference: "LAA-BBBBBB",
+               defendant_name: "Fred Bloggs",
+               firm_name: "Smith & (Partners) Ltd.")
+
+        create(:submission,
+               :with_pa_version,
+               laa_reference: "LAA-CCC123",
+               defendant_name: "Jimmy Buffer",
+               firm_name: "Über Legal Co.",
+               ufn: "311223/001")
+
+        create(:submission,
+               :with_pa_version,
+               laa_reference: "LAA-MiXeD1",
+               defendant_name: "Test Person",
+               firm_name: "Aardvark Smithson",
+               ufn: "123456")
+
+        create(:submission,
+               :with_pa_version,
+               laa_reference: "LAA-PUNC28",
+               defendant_name: "James O'Connor-Smith",
+               firm_name: "Legal & Law (International) Ltd.")
+      end
+
+      it "handles a completely invalid query" do
+        post search_endpoint, params: { query: "." }
+
+        expect(response).to have_http_status(:created)
+        expect(response.parsed_body["data"]).to match([])
+      end
+
+      it "ignores unmatched parentheses" do
+        post search_endpoint, params: { query: "Fred)" }
+
+        expect(response).to have_http_status(:created)
+        expect(response.parsed_body["data"].pluck("client_name")).to match(["Fred Bloggs", "Fred Arbor"])
+      end
+
+      it "ignores a real query attempt" do
+        post search_endpoint, params: { query: "LAA-AAAAAA)" }
+
+        expect(response).to have_http_status(:created)
+        expect(response.parsed_body["data"].pluck("laa_reference")).to match(%w[LAA-AAAAAA])
+      end
+
+      it "handles mixed case references" do
+        post search_endpoint, params: { query: "LAA-MiXeD1" }
+
+        expect(response).to have_http_status(:created)
+        expect(response.parsed_body["data"].pluck("laa_reference")).to match(%w[LAA-MiXeD1])
+      end
+
+      it "handles complex names" do
+        post search_endpoint, params: { query: "O'Connor-Smith" }
+
+        expect(response).to have_http_status(:created)
+        expect(response.parsed_body["data"].pluck("client_name")).to eq(["James O'Connor-Smith"])
+      end
+
+      it "ignores unescaped ampersands" do
+        post search_endpoint, params: { query: "Aardvark & Co" }
+
+        expect(response).to have_http_status(:created)
+        expect(response.parsed_body["data"].pluck("firm_name")).to match(["Aardvark & Co"])
+      end
+
+      it "doesn't treat ampersands as tokens" do
+        post search_endpoint, params: { query: "Aardvark &" }
+
+        expect(response).to have_http_status(:created)
+        expect(response.parsed_body["data"].pluck("firm_name")).to match(["Aardvark Smithson", "Aardvark & Co"])
+      end
+
+      it "handles records with multiple matches that have punctuation" do
+        post search_endpoint, params: { query: "Smith" }
+
+        expect(response).to have_http_status(:created)
+        expect(response.parsed_body["data"].pluck("firm_name")).to match(["Legal & Law (International) Ltd.", "Aardvark Smithson", "Smith & (Partners) Ltd."])
+      end
+
+      it "handles records that have punctuation and being in the query" do
+        post search_endpoint, params: { query: "Smith & (Partners) Ltd." }
+
+        expect(response).to have_http_status(:created)
+        expect(response.parsed_body["data"].pluck("firm_name")).to match(["Smith & (Partners) Ltd."])
+      end
+
+      it "handles UFNs" do
+        post search_endpoint, params: { query: "311223/001" }
+
+        expect(response).to have_http_status(:created)
+        expect(response.parsed_body["data"].pluck("firm_name")).to match(["Über Legal Co."])
+      end
+
+      it "handles 6 digit strings of numbers that could be UFNs" do
+        post search_endpoint, params: { query: "311223" }
+
+        expect(response).to have_http_status(:created)
+        expect(response.parsed_body["data"].pluck("firm_name")).to match(["Über Legal Co."])
+      end
+
+      it "does not handle 9 digit strings of numbers that could be UFNs" do
+        post search_endpoint, params: { query: "311223001" }
+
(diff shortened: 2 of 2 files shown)


robinsk1 (approved):

app/services/submissions/search_service.rb
```diff
@@ -20,6 +20,16 @@ def call
 
   private
 
+    def sanitize_query(query)
```
patrick-laa, line 23: Nit: It feels to me like maybe this belongs in Search, not SearchService, as it's really to do with the internals of how `where_terms` is conducted and the interpolation there.

app/services/submissions/search_service.rb
```diff
@@ -20,6 +20,16 @@ def call
...
+      return nil if query.blank?
+
+      query
+        .gsub(/[()]/, "")
```
patrick-laa, line 27: The initial issue seemed to be to do with a `&` in the query string, and that still seems to cause an error. Should we not guard against that as well? 

And rather than do this piecemeal for every naughty character we can think of, I wonder if it would make more sense to try to do something more comprehensive. https://stackoverflow.com/questions/14103880/escaping-special-characters-in-to-tsquery suggests `plainto_tsquery` or `$$` syntax?

spec/requests/submissions_search_spec.rb
```diff
@@ -640,5 +640,28 @@
         expect(laa_references).to match(%w[LAA-CCCCCC LAA-BBBBBB LAA-AAAAAA])
       end
     end
+
+    context "when searching for unsupported queries" do
+      before do
+        create_list(:submission, 2, :with_pa_version, defendant_name: "Joe Bloggs")
+
+        # This travel_to is used to test that "raw data" and [view] data are sync'd by page and order.
```
patrick-laa, line 648: Nit: this comment, and the accompanying block, has been copy-pasted from a test that actually used it, but it's nothing to do with what's being tested here. Can we get rid of everything in this `before` block we don't actually need?

app/services/submissions/search_service.rb
```diff
@@ -20,6 +20,16 @@ def call
...
+        .gsub(/[{}]/, "")
+        .gsub(/[\[\]]/, "")
```
ivanELEC, line 29: Another thing that might bite us in the back if we go down the route of simply removing the character is it will change the input search term. For example if someone types in "Chester&Noble", the search term will be "ChesterNoble" - whereas, it would yield better results as "Chester Noble" where the search terms are separate

app/services/submissions/search_service.rb
elken: Makes sense, just include here since this is where the bug was

app/services/submissions/search_service.rb
elken: So there's a couple of issues, one was unmatched parens, another was unescaped ampersands (which I'm looking at still, think I have it just adding another test case)

`plainto_tsquery` was a good find though, that seems to resolve it.

> can we also think about changing content? Whether we need to let the user know what the reason for the error is?

So all that we get back error-wise is "something is wrong with this query", we don't know exactly what; but using `plainto_tsquery` now it seems like nothing will cause it to break.

app/services/submissions/search_service.rb
elken: Ampersands aren't being removed, and the search is using partial matching anyway so it would still find it. See other comment anyway, this is redundant now.

app/services/submissions/search_service.rb
elken: The problem we have now though is anything the user enters we'll accept, so if they do add in an invalid character we don't have any way of discerning whether it should be part of the search query or not.

elken: Discussed internally and decided this is fine, and to ticket up how we handle invalid searches

app/models/search.rb
```diff
@@ -16,6 +16,7 @@ def self.where_terms(string)
         "#{clean_str}:A"
       elsif /\A\d+\/\d+\z/.match?(str) then "#{str}:A"
       elsif /\A\d+{6}\z/.match?(str) then "#{str}:*"
+      elsif /[&|!()*]/.match?(str) then "'''#{str.tr('/', '-')}''':B" # Escape special tsquery characters
```
patrick-laa, line 19: Rather than escape if there are special characters and not if there aren't, would it not make more sense to just escape every time? That way:

1) if there are other special characters we forgot about, they'll be automatically escaped
2) we reduce the possible edge cases. E.g. here you've gone for ":B" instead of ":*B", and that has consequences for doing partial matching, so you've created an edge case where if someone includes a partial match of a defendant name and one of these special characters the system will behave differently to if they  do just a partial search or just a special character.

So can I suggest just putting:

```
else
   "'''#{str.tr('/', '-')}''':*B"
end
```

elken: I tried that in testing and it failed so I assumed you couldn't partial match with this.

Odd, okay changed.

ivanELEC (approved):

spec/requests/submissions_search_spec.rb
```diff
@@ -640,5 +640,153 @@
...
+      end
+
+      it "handles complex names" do
+        post search_endpoint, params: { query: "O'Connor-Smith" }
+
+        expect(response).to have_http_status(:created)
+        expect(response.parsed_body["data"].pluck("client_name")).to eq(["James O'Connor-Smith"])
+      end
+
+      it "ignores unescaped ampersands" do
```
jogilder, line 714: this doesn't really check if it ignores the ampersand. I think you need another firm name like `aardvark smithson` in your dataset. That will only be returned if the & is ignored.

elken: Unescaped in the query, which is what one of the issues was; having an `&` in the query broke the search.

db/schema.rb
```diff
@@ -153,18 +153,37 @@
...
-      base.application_type,
-      base.version,
-      base.from_status,
-      base.from_time,
-      base.to_status,
-      base.to_time,
-      (base.from_time)::date AS from_date,
-      (base.to_time)::date AS to_date,
-      GREATEST(EXTRACT(epoch FROM (base.to_time - base.from_time)), (0)::numeric) AS processing_seconds
+   SELECT id,
```
jogilder, line 156: can we understand what is going on here please?

spec/requests/submissions_search_spec.rb
```diff
@@ -640,5 +640,153 @@
...
+      end
+
+      it "handles zero-width spaces" do
+        post search_endpoint, params: { query: "LAA-AB\u200BC" }
+
+        expect(response).to have_http_status(:created)
+        expect(response.parsed_body["data"]).to match([])
+      end
+
+      it "handles emojis" do
```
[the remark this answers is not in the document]
elken: It won't be entered by a user, but it proves that what we're doing to escape the search queries is working. If that degrades for any reason, we have a test to prove it.

db/schema.rb
elken: AFAIU nothing is changing, but I'll audit it when I've got the last bits of work done

jogilder (commented):
still not sure about taking out the - but can see it seems reasonable.

db/schema.rb
jogilder: But why are the semantics changed? They shouldn't be.

elken:
> still not sure about taking out the `-` but can see it seems reasonable.

I have changes to fix everything about to go up, if you want to jump on a call to run through it @jogilder let me know

db/schema.rb
elken: I don't know why, but I've removed them now anyway.

spec/requests/submissions_search_spec.rb
elken: That makes sense, have amended tests to hopefully demonstrate what's happening.

app/models/search.rb
```diff
@@ -10,17 +10,29 @@ class Search < ApplicationRecord
...
-      elsif /\A\d+{6}\z/.match?(str) then "#{str}:*"
+      # We escape with single quotes
+      str = str.gsub("'", "''")
+
+      # Is this an LAA reference or a UFN? (LAA-123ABC or 311223/001)
+      if str.start_with?("laa-") || /\A\d+\/\d+\z/.match?(str) then %('''#{str}''':A)
+
+      # Is this MAYBE a UFN? (6 digits for the first part, 9 digits for the full)
+      elsif /\A\d{6}\z/.match?(str) then "#{str}:*"
+      # Don't know what it is, but escape it and give it a lower weight anyway
```
patrick-laa, line 23: Nit: This is a slightly misleading comment. `B` is not a "lower" weight than `A`. Higher/lower doesn't really come into play for weights until you start messing with `ts_rank`. Rather, `A` is the "LAA reference and UFN" group, while `B` is the  "Defendant name and firm name" group. So we're saying "we know this must be either a defendant or first name, so we escape it and mark it as a query against the "B" weight"

app/models/search.rb
```diff
@@ -10,17 +10,29 @@ class Search < ApplicationRecord
...
       end
     end
 
+    # The way the lexemes are setup in the search vector means a typical record ends up looking like
+    #
+    # '-123456':8A '010124/001':6A 'laa-123456':10A 'jason':2B 'jason/jim':1B 'jim':3B  'read':4B,5B
+    #
+    # In which we have the expected fields converted to lexemes, however in the case of the "Jason/Jim" first name; we generate lexemes for both the full and partial matches of both names.
+    #
+    # We only do this for defendant name as caseworkers use that syntax for alternate names.
```
patrick-laa, line 35: Nit: Are you saying that providers enter defendant names using "/" to indicate alternate names? If so, I can't see any evidence of that behaviour in our DB and don't remember any requirements around that. Or are you saying caseworkers _search_ for defendant names using "/"? In which case, shouldn't we have tests to verify that that works as expected in search?

app/models/search.rb
elken: Yeah I wrote that before I understood what they were for better

app/models/search.rb
elken: It was definitely raised at some point, but if it turns out that it's not actually a thing anymore that'll be very annoying as the bulk of the PR now is making that work...

And yeah there's already tests for alternate names, https://github.com/ministryofjustice/laa-crime-application-store/blob/main/spec/models/search_spec.rb#L178

patrick-laa: So those tests for alternate names specify that searching for "Jason/Jim" should only match defendants actually called "Jason/Jim" in the DB. In which case we shouldn't need to generate lexemes for "generate lexemes for both the full and partial matches of both names" should we? I think the goal of those tests was to demonstrate "the search doesn't break if there's a slash in the query", nothing more complex than that.

elken: Ah I think you're right, that's annoying.

I must have thankfully misremembered the full requirement, I'll remove the extra lexeme creation bits then.

elken: Wait one of those defendant names in the test is "Jason/Jim Read" @patrick-laa 

Does this mean the test is wrong or the requirement?

elken: We can't search for that though, we can only match either the whole token or the start of the token. I'm not sure why it was working _before_, but now that we have to escape the search query `/` is taken literally so while it doesn't break the query we can't use it without generating the other lexemes.

patrick-laa: Ah but first name and last name are separate tokens. This is why if the defendant is called "Jim Beam" you can search for "Beam" as well as searching for "Jim" - we shove first name, last name and firm name _separately_ into `search_fields`, so you can search on any one of them.

elken: You can, but when using `to_tsquery` raw as we were before, `-` is parsed as a word boundary so the `*` (match from start, _NOT_ partial matching) was able to work on both James and Jim.

Now we escape the query, special characters are no longer treated as word boundaries so all we can do is search the whole token or from the start of a token, making the "Jim" bit in "Jason/Jim" never resolve.

patrick-laa: But I have the test passing with your escaping but without any new extra lexeme code. The existing code replaces `/` with `-` in defendant names in the `search_fields` virtual column, and replaces `/` with `-` in the query. So `Json/Jim Read`in the DB becomes `json-jim:1B read:4B` in the virtual column, and `Jason/Jim` becomes `jason-jim` in the search query, and it just works. We don't need to add anything more to keep the tests passing.

patrick-laa: @elken aah, I finally understand the source of the confusion: That's because the `down` bit of your migration is buggy. If I apply your migration and then roll it back, I too end up with a bunch of failing tests. BUT I also end up with a schema that no longer resembles what it did before I first applied the migration. Whereas if instead I simply don't apply your migration in the first place and simply use your new code and new tests... they pass.

You've been led astray by a dodgy rollback!

elken: I'm afraid I'm not observing that either. With a fresh DB and no migration from this PR, all the LAA reference tests fail as the initial migration removes the `-` and now we don't do that.

patrick-laa: So look: All I'm doing is adding the escaping that you discovered, and all the new test cases you added, and all the James/Jim stuff still passes - https://github.com/ministryofjustice/laa-crime-application-store/pull/314/files

elken: @patrick-laa after our discussion and the latest changes are we happy with this now?

patrick-laa: So I feel like I should no longer be the arbiter of approval, but to me there's 3 things I'd consider changing:

1. The Ruby comment that this GH comment thread is attached to is inaccurate
2. You don't really need the clause on line 29-31 because you're still doing a tr on line 36 and we know that that tr already does everything that the lines 30-31 does, functionally speaking. 
3. I know that the "search for 'B' matches 'A/B' functionality isn't strictly needed, but since we already have that functionality and it already works, I don't think we really need to take the tests that exercise that functionality out.

patrick-laa: 1. "caseworkers use that syntax for alternate names" - I'm not sure we have any evidence that they do
2. Without seeing the code from that CI run I can't comment on what went wrong for you, but look, here it is working without that extra clause: https://app.circleci.com/pipelines/github/ministryofjustice/laa-crime-application-store/1435/workflows/23f78080-c633-4c7c-b14a-382fed0e962b/jobs/9672
3. The test doesn't make an incorrect assumption about the system. It demonstrates that if the query is "Jim", and the defendant has been entered by the provider as having first name "Jason/Jim", searching for "Jim" wlll return the record. Which is true (hence why the test passes). It accurately describes the behaviour of the system. The fact that we don't need it to behave that way doesn't mean we want our tests to _less accurately_ describe how the system behaves.

elken: 1. Oh you're right
2. I removed the lines you mentioned which are needed for the tests removed by point 3, but this is made moot if we just keep the tests around
3. I don't agree here, but keeping the tests renders point 2 moot so I'll just leave them

patrick-laa (approved):

PR #313, opened 2024-10-31
CRM457-2176: Update view definition for submission_by_date to handle CRM4 resubmissions

ivanELEC:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2176)

### DIFF
--- a/spec/postgres_views/submissions_by_date_spec.rb
+++ b/spec/postgres_views/submissions_by_date_spec.rb
@@ -7,7 +7,7 @@ RSpec.describe "submissions_by_date" do
     end
   end
 
-  context "application type is crm7" do
+  context "when application type is crm7" do
     let(:application_type) { "crm7" }
 
     it "reports submissions and resubmissions for a given date" do
@@ -63,7 +63,7 @@ RSpec.describe "submissions_by_date" do
     end
   end
 
-  context "application type is crm4" do
+  context "when application type is crm4" do
     let(:application_type) { "crm4" }
 
     it "reports submissions and resubmissions for a given date" do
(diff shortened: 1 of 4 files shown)


spec/postgres_views/submissions_by_date_spec.rb
```diff
@@ -7,38 +7,115 @@
...
+          { event_type: "provider_updated", submission_version: 2, created_at: 1.day.ago },
+          { event_type: "provider_updated", submission_version: 3, created_at: 1.day.ago },
+        ],
+      )
+      expect(klass.all.map(&:attributes)).to eq([
+        { "event_on" => 1.day.ago.to_date, "application_type" => "crm4", "submission" => 1, "resubmission" => 2, "total" => 3 },
+      ])
+    end
+
+    it "does not use crm4 criteria for resubmissions" do
```
jogilder, line 107: Is this really the right way of describing this test and the one above? I guess this is `does not use `new_version` for resubmissions and the one in crm7 is that it does not use `provider_updated` for resubmission?

jogilder: and would you ever end up with this situation?

ivanELEC: I feel like a less literal description and just highlighting that this is the criteria for either type of application makes the intention clearer - in the future, if the criteria changes for whatever reason, we can just update the content of the test (particuarly because the test can get wordy - "does not use new_version with a submission_version higher than 1 for resubmission"

This wouldn't happen currently of course - but doesn't mean we shouldn't test for it. The source of the data is events so if we change how our events works and have to update this view, we then have a way of checking whether the change breaks the view or not

jogilder: ok but you still can't call this one does not use crm4 criteria for resubmissions because it does.

ivanELEC: oooooh sorry my mistake, I completely missed the fact that it said crm4

jogilder (approved):
LGTM

PR #315, opened 2024-11-05
CRM457-2216: Everything needed to load all lists from the app store

patrick-laa:
## Description of change
As part of the progressive move towards removing the caseworker DB, add functionality to the app store to load all list UIs from the app store rather than the local db:
- Search on currently assigned user ("Your claims/applications") and endpoints to keep this in sync
- Search on multiple statuses ("Open/closed claims/applications")
- Search on account number ("Related applications")

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2216)


elken (commented):
I'm happy with this, but I think for both this and the other PR it should be reviewed by everyone.

PR #325, opened 2024-11-12
CRM457-2216 part 6: App store handles new_version, autogrant and expiry

patrick-laa:
## Description of change
So that we don't have to rely on the caseworker app DB or the caseworker app being pushed new data as soon as it comes in

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2216)


spec/requests/update_submission_spec.rb
```diff
@@ -130,4 +130,17 @@
     patch "/v1/submissions/#{submission.id}", params: { application_state: "granted", application: { new: :data }, json_schema_version: 1 }
     expect(submission.ordered_submission_versions.find_by(id: pending_version.id)).to be_nil
   end
+
+  context "when provider is updating" do
+    before { allow(Tokens::VerificationService).to receive(:call).and_return(valid: true, role: :provider) }
+
+    it "adds a new version event if appropriate" do
```
elken, line 137: When wouldn't it be appropriate? Should we add a case for that?

elken (approved):

PR #328, opened 2024-11-14
CRM-457 2154 Add 'high value' filter to search

imtiazAhmd:
## Description of change

Adds High value to the search

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2154)


db/schema.rb
```diff
@@ -62,6 +70,64 @@
 
   add_foreign_key "application_version", "application", name: "application_version_application_id_fkey"
 
+  create_view "geography_columns", sql_definition: <<-SQL
```
robinsk1, line 73: not too sure why these are included here for this change and why we would need a view of Postgis extension tables?

imtiazAhmd: No idea, this was auto generated when I did the migration.

patrick-laa: So 2 things have happened. First, we enabled postgis so that LocationService can figure out if a postcode is inside the M25 (because this now happens here rather than in assess). Second, one of our Rails dependencies is now struggling with postgis, because it started doing this on assess recently to - when the schema file is regenerated it's seeing some postgis tables and thinking they're manually created tables that belong in the schema file. (Which then messes up when you try to use the schema to run things like `rails db:prepare`

db/views/searches_v07.sql
```diff
@@ -0,0 +1,45 @@
...
+      ON app.id = app_ver.application_id AND app.current_version = app_ver.version
+)
+SELECT
+  app.id,
+  app_ver.id as application_version_id,
+  app_ver.application ->> 'laa_reference' as laa_reference,
+  app_ver.application -> 'firm_office' ->> 'name' as firm_name,
+  app_ver.application -> 'firm_office' ->> 'account_number' as account_number,
+  app_ver.application ->> 'service_name' as service_name,
+  (app_ver.application -> 'cost_summary' ->> 'high_value')::boolean as high_value,
```
patrick-laa, line 24: The problem with this is that the `high_value` flag has only recently started being set, and it wasn't backfilled, cos I wasn't sure if we'd end up using it. But I think it now makes sense to backfill it, and the good news is it's quite a simple migration (which we only need to do in the app store, because the CW app is going away):

For all submission versions tied to a CRM7 submission where `cost_summary->high_value` is null but `cost_summary` is not null, update to add a high value flag, where the flag is set according to whether `(cost_summary->profit_costs->>gross_cost)::decimal >= 5000`.

For all submission versions tied to a CRM7 submission where `cost_summary` is null, it's harder, because you'll need to create a cost summary hash, and to pass the schema validator, the cost_summary will need to have a bunch of required fields, but we don't know the value of those fields. But that's OK cos we never read them, so you'll want to put in something like this:

```
"cost_summary":{
        "disbursements":{
          "gross_cost":"-1",
          "net_cost":"-1",
          "vat":"-1"
        },
        "profit_costs":{
          "gross_cost":"-1",
          "net_cost":"-1",
          "vat":"-1"
        },
        "travel":{
          "gross_cost":"-1",
          "net_cost":"-1",
          "vat":"-1"
        },
        "waiting":{
          "gross_cost":"-1",
          "net_cost":"-1",
          "vat":"-1"
        },
        "high_value": <whether or not the claim is flagged as high risk>
      }
      
      It's not perfect, in that high risk is not quite the same thing as high value, but it's good enough - that's applying the same fallback that our auto-assignment uses.

File: spec/requests/submissions_search_spec.rb [hunk 3]
```diff
@@ -479,6 +479,31 @@
...
+      before do
+        create(:submission, :with_nsm_version,
+               high_value: true,
+               laa_reference: "LAA-AAAAA1")
+
+        create(:submission, :with_nsm_version,
+               high_value: true,
+               laa_reference: "LAA-AAAAA2")
+
+        create(:submission, :with_nsm_version,
```
patrick-laa, line 492: Worth making this explicitly `high_value: false` so you can prove you're correctly casting to a boolean?

app/services/submissions/search_service.rb
```diff
@@ -28,6 +28,7 @@ def search_query
       relation = relation.where("has_been_assigned_to ? :caseworker_id", caseworker_id:) if caseworker_id
       relation = relation.where(assigned_user_id:) if assigned_user_id
       relation = relation.where(account_number:) if account_number
+      relation = relation.where(high_value:) if high_value
```
patrick-laa, line 31: This works for the current requirement, but it's misleading. As it stands it makes it impossible to filter by NOT high value, because if `search_params[:high_value]` is set to false, then `if high_value` will prevent this line from running. You'd be better off with `unless high_value.nil?`

lib/tasks/fixes.rake
```diff
@@ -160,6 +160,35 @@ namespace :fixes do
...
   end
 
+  desc "Append high_value flag to submissions"
+  task adds_high_value: :environment do
+    # Query to select all submission versions that doesn't have a high_value flag
+    # and go through each submission version to add the high_value flag
+    # and show logs for total affected submission versions
+    # and show logs for successful submission versions
+    # and show logs for failed submission versions
+    submission_versions = SubmissionVersion.where("application -> 'cost_summary' ->> 'high_value' IS NULL")
```
patrick-laa, line 170: This doesn't exclude PA submissions - you definitely want to exclude them

lib/tasks/fixes.rake
```diff
@@ -160,6 +160,35 @@ namespace :fixes do
...
+    total_affected_submission_versions = submission_versions.count
+    successful_submission_versions = 0
+    failed_submission_versions = 0
+    submission_versions.each do |submission_version|
```
patrick-laa, line 174: It's more performant to do `find_each` than `each`, because the former batch loads them from the DB rather than pulling them all into memory from a single query

lib/tasks/fixes.rake
```diff
@@ -160,6 +160,35 @@ namespace :fixes do
...
+      if submission_version.application['cost_summary']['profit_costs']['gross_cost'].to_f >= 5000
```
patrick-laa, line 175: This will explode when `cost_summary` is null, as it is for cases that pre-date having cost_summary in the blob

lib/tasks/fixes.rake
```diff
@@ -160,6 +160,35 @@ namespace :fixes do
...
+        submission_version.application['cost_summary']['high_value'] = true
+      else
+        submission_version.application['cost_summary']['high_value'] = false
+      end
+
+      if submission_version.save!(touch: false)
```
patrick-laa, line 181: This if block is redundant - if `save!` fails it won't return false, it will raise an error, blowing up your entire rake task.

patrick-laa (commented):
Given there's quite a lot of little issues with that rake task that I've seen (and therefore might be more I've not seen), it might be best to add some tests to exercise it before we try it in production.

db/views/searches_v07.sql
```diff
@@ -22,6 +22,7 @@ SELECT
   app_ver.application -> 'firm_office' ->> 'name' as firm_name,
   app_ver.application -> 'firm_office' ->> 'account_number' as account_number,
   app_ver.application ->> 'service_name' as service_name,
+  (app_ver.application -> 'cost_summary' ->> 'high_value')::boolean as high_value,
```
patrick-laa, line 25: Let's remove this line - you don't want to be modifying a previous version definition of the view

db/views/searches_v09.sql
```diff
@@ -0,0 +1,47 @@
...
+)
+SELECT
+  app.id,
+  app_ver.id as application_version_id,
+  app_ver.application ->> 'ufn' as ufn,
+  app_ver.application ->> 'laa_reference' as laa_reference,
+  app_ver.application -> 'firm_office' ->> 'name' as firm_name,
+  app_ver.application -> 'firm_office' ->> 'account_number' as account_number,
+  app_ver.application ->> 'service_name' as service_name,
+  (app_ver.application -> 'cost_summary' ->> 'high_value')::boolean as high_value,
```
patrick-laa, line 25: There's a bit of double-casting here - you're taking `high_value`, which is a boolean, casting it to text with `->>`, and then casting it back to a boolean. Can we avoid this? (I suppose it might be that this is the only way to ensure the view doesn't explode if high_value turns out to have an unexpected type)

lib/tasks/CRM457_2154.rake
```diff
@@ -0,0 +1,47 @@
+# https://dsdmoj.atlassian.net/browse/CRM457-2154
+
+namespace :CRM457_2154 do
+  desc "Append high_value flag to submissions"
+  task adds_high_value: :environment do
+    # Query to select all submissions that don't have a high_value flag
+    # and go through each submission version to add the high_value flag:
+    # use cost_summary's gross_cost if present, otherwise use submission's application risk
+    submissions = Submission.where(application_type: 'crm7')
+                            .joins(:ordered_submission_versions)
```
patrick-laa, line 10: You could simplify this logic slightly if you switched it round:

versions = SubmissionVersion.includes(:submission).joins(:submission).where("application_version.application -> 'cost_summary' ->> 'high_value' IS NULL AND application.application_type = 'CRM7'"

That way you avoid a nested `each` loop and do fewer DB lookups

patrick-laa (approved):

PR #331, opened 2024-11-19
Add pooling queue to Sidekiq config

robinsk1:
## Description of change

It looks like we've added a new queue for the expire application cron job but we're not telling our workers about it

### DIFF
--- a/config/sidekiq.yml
+++ b/config/sidekiq.yml
@@ -2,3 +2,4 @@
 :logfile: ./log/sidekiq.log
 :queues:
   - default
+  - pooling


ivanELEC (approved):

config/sidekiq.yml
```diff
@@ -2,3 +2,4 @@
 :logfile: ./log/sidekiq.log
 :queues:
   - default
+  - pooling
```
patrick-laa, line 5: I don't think this is doing what you think it's doing

patrick-laa: Oh, my bad, it does exactly what you think it's doing, I was misreading. Sorry!

patrick-laa (approved):

PR #334, opened 2024-11-26
CRM457-2221: Allow providers to generate filtered lists

patrick-laa:
## Description of change
- Providers can access the search endpoint
- The search endpoint allows passing in multiple office codes
- Searches can be sorted by UFN
- Searches can be sorted by the date of the most recent state change (e.g. for submitted applications it's when they were submitted, not the date of the most recent caseworker assignment)
- Caseworker app can no longer subscribe/unsubscribe (since it doesn't do that any more)

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2221)

### DIFF
--- a/app/services/submissions/search_service.rb
+++ b/app/services/submissions/search_service.rb
@@ -1,6 +1,17 @@
 module Submissions
   class SearchService
-    SORTABLE_COLUMNS = %w[laa_reference firm_name client_name last_updated status_with_assignment risk_level service_name].freeze
+    SORTABLE_COLUMNS = %w[
+      ufn
+      laa_reference
+      firm_name
+      client_name
+      last_updated
+      status_with_assignment
+      risk_level
+      service_name
+      account_number
+      last_state_change
+    ].freeze
 
     attr_reader :search_params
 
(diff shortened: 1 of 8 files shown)


elken (approved):

app/services/submissions/search_service.rb
```diff
@@ -1,6 +1,17 @@
 module Submissions
   class SearchService
-    SORTABLE_COLUMNS = %w[laa_reference firm_name client_name last_updated status_with_assignment risk_level service_name].freeze
+    SORTABLE_COLUMNS = %w[
```
elken, line 3: Doesn't directly relate to this ticket but having these fields in the common gem will help us with the future query param validation ticket, agreed?

ivanELEC (approved):

app/services/submissions/search_service.rb
patrick-laa: Agreed. I think there's a question of how much search-related stuff goes into the common gem, because a lot of this stuff knows to a certain extent about the structure of the NSM and PA payloads, but trying to move e.g. view definitions into the gem is gonna send us loopy...

elken: Yeah I think a sensible drawing of that line is views, and any code that is identical in both (hence, _common_). If we start going down the route of adding too much conditional logic to the gem it ends up being a dumping ground rather than a neat way to make things more DRY

PR #338, opened 2024-12-05
CRM457-2322 part 1: Pull anonymised records from prod to localhost

patrick-laa:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2322)

Note that this is a first step towards pulling records from prod into a secure pre-prod environment, but it delivers incremental usefulness as a standalone step.

### DIFF
--- a/bin/download_anonymised
+++ b/bin/download_anonymised
@@ -1,11 +1,14 @@
 #!/bin/sh
+set -e
 
 LAA_REF=$1
 
-kubectl exec deploy/laa-crime-application-store-app \
+echo "Downloading anonymised record..."
+OUTPUT=$(kubectl exec deploy/laa-crime-application-store-app \
 -it -n laa-crime-application-store-production -- \
-bundle exec rails anonymised:download\[$LAA_REF\] > temp-anonymised.txt
+bundle exec rails anonymised:download\[$LAA_REF\])
 
-bundle exec rails anonymised:import\[temp-anonymised.txt\]
+echo "Importing anonymised record..."
+ANONYMISED_DOWNLOAD_OUTPUT=$OUTPUT bundle exec rails anonymised:import\[ANONYMISED_DOWNLOAD_OUTPUT\]
 
-rm temp-anonymised.txt
+echo "Done."
--- a/bin/download_anonymised
+++ b/bin/download_anonymised
@@ -1,0 +1,11 @@
+#!/bin/sh
+
+LAA_REF=$1
+
+kubectl exec deploy/laa-crime-application-store-app \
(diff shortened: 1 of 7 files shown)


lib/tasks/anonymised.rake
```diff
@@ -0,0 +1,50 @@
...
+  task :download, [:laa_reference] => :environment do |_, args|
+    submission = Submission.joins(:ordered_submission_versions)
+                           .find_by("application_version.application->>'laa_reference' = ?",
+                                    args[:laa_reference])
+    service_type = submission.application_type == 'crm7' ? :nsm : :prior_authority
+    anonymised = LaaCrimeFormsCommon::Anonymiser.anonymise(
+      service_type,
+      submission.latest_version.application
+    )
+    puts "JSONSTART"
```
elken, line 12: What's this for? I see you're using it to split later down but I don't understand why you can't just return JSON.

patrick-laa: So because I'm getting the content by just `> file`, unfortunately the file captures all output of the command, including the lines that get spat out on startup like `I, [2024-12-06T15:29:33.631366 #566]  INFO -- : [PrometheusExporter] Starting server....`. I needed to delineate between that and the actual JSON content.

There's probably a smarter way...

bin/download_anonymised
```diff
@@ -0,0 +1,11 @@
...
+
+LAA_REF=$1
+
+kubectl exec deploy/laa-crime-application-store-app \
+-it -n laa-crime-application-store-production -- \
+bundle exec rails anonymised:download\[$LAA_REF\] > temp-anonymised.txt
+
+bundle exec rails anonymised:import\[temp-anonymised.txt\]
+
+rm temp-anonymised.txt
```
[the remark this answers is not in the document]
patrick-laa: That's a good idea. I was omitting `set -e` for a similar result, but it did mean that if the download failed the import would still try to run, so I'd get 2 failures for the price of 1

lib/tasks/anonymised.rake
elken: Oh right I see.

What if we split out the logic that does the downloading and then for print we just display the data we pulled and for the other task we just add it to the database and do away with the temp file altogether?

patrick-laa: I'm not sure I follow. I can write tempfile out of the equation by pushing the output from the first command to the input of the second, but I'm still only getting the data from the remote pod to my local machine via stdout, which is "polluted" by all the _other_ stuff that gets written to it by triggering the rake task.

PR #339, opened 2024-12-06
CRM457-2221: Remove redundant pub-sub mechanism

patrick-laa:
## Description of change
Now that neither provider nor caseworker relies on pub/sub we can remove the mechanism entirely.

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2221)


elken (approved):
I love seeing so many removed lines.

I'm sure you are; but are you sure this is all the places where we do pub-sub in the app store? Approved based on the assumption you have.

PR #340, opened 2024-12-09
CRM457-2322 part 2: Push record to UAT

patrick-laa:
## Description of change
- Move anonymised records to UAT instead of localhost
- Allow deleting from UAT by rake task

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2322)


elken (changes requested):
I don't think we should _replace_ the existing download tool, instead maybe handle optional arguments (and default to local, so we don't accidentally spam UAT with claims)

bin/delete_anonymised
```diff
@@ -1,5 +1,12 @@
 #!/bin/sh
 
 LAA_REF=$1
+DESTINATION_ENVIRONMENT=$2
 
-bundle exec rails anonymised:delete\[$LAA_REF\]
+if [[ $DESTINATION_ENVIRONMENT == "local" ]]; then
+  bundle exec rails anonymised:delete\[$LAA_REF\]
```
elken, line 7: So we can just do `if != local` here and put the `kubectl` command in that as both conditions run the same line

bin/download_anonymised
```diff
@@ -2,13 +2,21 @@
...
-echo "Importing anonymised record..."
-ANONYMISED_DOWNLOAD_OUTPUT=$OUTPUT bundle exec rails anonymised:import\[ANONYMISED_DOWNLOAD_OUTPUT\]
+echo "Importing anonymised record to $2..."
+
+if [[ $DESTINATION_ENVIRONMENT == "local" ]]; then
+  ANONYMISED_DOWNLOAD_OUTPUT=$OUTPUT bundle exec rails anonymised:import\[ANONYMISED_DOWNLOAD_OUTPUT\]
+else
+  kubectl exec deploy/laa-crime-application-store-app \
+  -it -n laa-crime-application-store-$DESTINATION_ENVIRONMENT -- \
+  env ANONYMISED_DOWNLOAD_OUTPUT="${OUTPUT}" bundle exec rails anonymised:import\[ANONYMISED_DOWNLOAD_OUTPUT\]
```
elken, line 19: Same here, `env` isn't needed but the shell escape quoting should be

bin/delete_anonymised
patrick-laa: Ah but the kubectl command and the `bundle exec rails` aren't separate commands, they're the same one with a line break. I don't think you can if-block a partial statement

elken: Oh I see the line breaks now, yeah fine

bin/download_anonymised
patrick-laa: FWIW env really is needed - without it you get `error: Internal error occurred: error executing command in container: failed to exec in container: failed to start exec...`. You may not have spotted that line 19 is not a standalone statement, it's a continuation of the statement starting on line 17, and is therefore  run in the kube pod and needs slightly special escaping to get it there.

elken: Yeah line break again, it is. Line 15 should have the same shell escape quoting though

elken (approved):

bin/download_anonymised
```diff
@@ -16,7 +16,7 @@ if [[ $DESTINATION_ENVIRONMENT == "local" ]]; then
 else
   kubectl exec deploy/laa-crime-application-store-app \
   -it -n laa-crime-application-store-$DESTINATION_ENVIRONMENT -- \
-  env ANONYMISED_DOWNLOAD_OUTPUT="${OUTPUT}" bundle exec rails anonymised:import\[ANONYMISED_DOWNLOAD_OUTPUT\]
+  env ANONYMISED_DOWNLOAD_OUTPUT="$OUTPUT" bundle exec rails anonymised:import\[ANONYMISED_DOWNLOAD_OUTPUT\]
```
elken, line 19: That's the opposite of what I meant :P Escaping it prevents something in the output causing it to be interpreted as part of the script, so you DO want them both to be `"${OUTPUT}"`

PR #349, opened 2024-12-13
CRM457-2347: Correct the mailer to always send `to` as an array

elken:
## Description of change

Wrap the mailer to ensure that we always send `to` as an array even if it's invalid.

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2347)

### DIFF
--- a/app/mailers/email_to_provider_mailer.rb
+++ b/app/mailers/email_to_provider_mailer.rb
@@ -5,7 +5,7 @@ class EmailToProviderMailer < GovukNotifyRails::Mailer
       message = message_class.new(submission.latest_version.application)
       set_template(message.template)
       set_personalisation(**message.contents)
-      mail(to: message.recipient)
+      mail(to: Array(message.recipient))
     end
   end
   # :nocov:
--- a/config/initializers/govuk_notify.rb
+++ b/config/initializers/govuk_notify.rb
@@ -1,2 +1,14 @@
 ActionMailer::Base.add_delivery_method :govuk_notify, GovukNotifyRails::Delivery,
-                                       api_key: ENV.fetch('GOVUK_NOTIFY_API_KEY', nil)
+                                       api_key: ENV.fetch("GOVUK_NOTIFY_API_KEY", nil)
+
+class MailDeliveryJobWrapper < ActionMailer::MailDeliveryJob
+  def serialize
+    super.tap do |args|
+      args["arguments"][3] = Array(args["arguments"][3])
+    end
+  end
+end
+
+# Set this so we ensure that the `to` list we pass to
+# govuk_notify_rails is always an array
+Rails.application.config.action_mailer.delivery_job = "MailDeliveryJobWrapper"
--- a/spec/jobs/mail_delivery_job_wrapper_spec.rb
+++ b/spec/jobs/mail_delivery_job_wrapper_spec.rb
@@ -1,0 +1,37 @@
+require "rails_helper"
+
+# rubocop:disable Rails/ApplicationMailer, Rails/I18nLocaleTexts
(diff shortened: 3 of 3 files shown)


app/mailers/email_to_provider_mailer.rb
```diff
@@ -5,7 +5,7 @@ def notify(message_class, submission)
       message = message_class.new(submission.latest_version.application)
       set_template(message.template)
       set_personalisation(**message.contents)
-      mail(to: message.recipient)
+      mail(to: Array(message.recipient))
```
robinsk1, line 8: maybe instead of just casting we could have a class whose responsibility is to clean up a message for  notify_gov_uk interface expectations - might signal intent/be testable also

config/initializers/govuk_notify.rb
```diff
@@ -1,2 +1,14 @@
 ActionMailer::Base.add_delivery_method :govuk_notify, GovukNotifyRails::Delivery,
-                                       api_key: ENV.fetch('GOVUK_NOTIFY_API_KEY', nil)
+                                       api_key: ENV.fetch("GOVUK_NOTIFY_API_KEY", nil)
+
+class MailDeliveryJobWrapper < ActionMailer::MailDeliveryJob
+  def serialize
+    super.tap do |args|
+      args["arguments"][3] = Array(args["arguments"][3])
```
robinsk1, line 7: is this just a catch-all?

app/mailers/email_to_provider_mailer.rb
elken: Everything else matches already AFAICT

config/initializers/govuk_notify.rb
patrick-laa, line 7: This feels a little bit nervous-making - are we sure that `args["arguments"][3]` is always going to be the `to` field? The fact that we now have this string-to-array issue covered nicely in `EmailToProviderMailer` makes me feel like this wrapper is possibly more of a liability than a win.

app/mailers/email_to_provider_mailer.rb
patrick-laa: You could argue that the place to do this is inside the `message.recipient` definition in the common gem. The advantage of that would be that since both the app store and the caseworker app use the common gem you'd only need to make this change once; otherwise you have to do this both here and in Nsm::EmailToProviderMailer and PriorAuthority::EmailToProviderMailer in the caseworker app.

N.B. either way, you also need to make the change in Nsm::SubmissionMailer and PriorAuthority::SubmissionMailer, as they don't use the same `message` system to send their acknowledgement-of-submission emails.

patrick-laa (commented):
Approving this on the grounds that if you're confident this is how you want to go about it it's not an unreasonable approach, but I've flagged some alternatives

config/initializers/govuk_notify.rb
robinsk1: yeah, basing it on an index feels brittle - might have string-to-arraying higher up the chain?

app/mailers/email_to_provider_mailer.rb
elken: That's a good point I wasn't aware of until our discussion yesterday. Okay I'll shelve this for now and include it as part of the "move where we send emails" ticket

config/initializers/govuk_notify.rb
elken: The test always passes, so I'm confident enough to say yes it is. It is a bit scary though

elken:
Closing to be resolved as part of https://dsdmoj.atlassian.net/browse/CRM457-2354 instead

PR #350, opened 2024-12-13
CRM457-2217: Stop looking at the events column

patrick-laa:
## Description of change
The `events` column should _only_ be used to populate the 'History' UIs in the caseworker app. So:
- Rewrite views we need to stop using the events column
- Remove views we don't need that use the events column
- Remove the virtual column that references the events column
- Rename the events column to `caseworker_history_events`
- Bonus: Clear out a bunch of one-off "fix" code that is no longer relevant, some of which dug heavily into the old events stuff

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2217)


db/migrate/20240806130123_backfill_disbursement_position.rb
```diff
@@ -14,13 +14,13 @@ def up
     # OR submission version with any disbursements
     # SubmissionVersion.where("application ->'disbursements' IS NOT NULL")
 
-    claims = Submission.where(application_type: "crm7")
+    # claims = Submission.where(application_type: "crm7")
```
elken, line 17: Why not removed?

elken (approved):

db/migrate/20240806130123_backfill_disbursement_position.rb
patrick-laa: I had a notion it would be helpful to see what the migration _used_ to do, but the more I think about it the more that's nonsense. I'll remove this (and the one below)

elken: There is some logic to maybe keeping the empty migrations around, I don't know how Rails handles "missing" migrations. I assume it handles cleanly and if so probably best just removing them yeah

robinsk1 (approved):

PR #364, opened 2025-01-21
[CRM457-1792] GDPR scheduled job for attachments 

robinsk1:
## Description of change

- [x] S3 setup dev CP
- [x] S3 setup uat CP
- [x] S3 setup Prod CP
- [x] dev testing
- [x] common gem schema changes https://github.com/ministryofjustice/laa-crime-forms-common/pull/37

This change is scoped to reviewed claims only - combines filter and delete ACs for GDPR qualifying reviewed NSM claims

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1792) (https://dsdmoj.atlassian.net/browse/CRM457-2374)

i've outlined suggested manual testing in comments on ticket https://dsdmoj.atlassian.net/jira/software/c/projects/CRM457/boards/1192?selectedIssue=CRM457-1792


app/jobs/nsm/schedule_delete_reviewed_docs.rb
```diff
@@ -0,0 +1,24 @@
...
+      return if filtered_claims.empty?
+
+      filtered_claims.each do |claim|
+        Nsm::DeleteReviewedClaimDocs.perform_later(claim.id)
+      end
+    end
+
+    def filtered_claims
+      state = %w[granted part_grant rejected expired]
+
```
ivanELEC, line 15: maybe comments here would be good to express the sentiment of what you want this query to be?

app/jobs/nsm/schedule_delete_reviewed_docs.rb
```diff
@@ -0,0 +1,24 @@
...
+      Submission.joins("LEFT JOIN application_version latest_version ON latest_version.application_id = application.id AND
+        latest_version.version = (SELECT MAX(version)
+        FROM application_version all_versions
+        WHERE all_versions.application_id = application.id)")
+        .where(state:, application_type: "crm7", updated_at: ..6.months.ago)
```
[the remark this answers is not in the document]
elken: It should also be `6.months.ago.end_of_day` so that we don't miss by a day

app/jobs/nsm/delete_reviewed_claim_docs.rb
```diff
@@ -0,0 +1,72 @@
...
+        if file_uploader.destroy(file_path)
+          Rails.logger.info "Deleted #{file_path} for submission #{submission.id}"
+        else
+          Rails.logger.error "Delete failed for document: #{file_path} submission: #{submission.id}"
+          raise GDPRDeletionError, "Failed delete document: #{file_path} submission: #{submission.id}"
+        end
+      end
+
+      submission.with_lock do
+        submission.current_version += 1
```
ivanELEC, line 38: One thing to note is that our internal team metrics dashboards rely on version increments only being based on state changes - I think it's probably correct to increment the version here as we are changing application data and there are probably some (admittedly a bit complex) ways around this in the metrics queries, but I'm thinking...the important piece of information we are changing is there any reason why we couldn't store that at the top level of the submission instead of in the application data payload? That way we could avoid having to create a new version?

ivanELEC (commented):
left some comments to discuss

app/jobs/nsm/delete_reviewed_claim_docs.rb
robinsk1: i guess having that flag in the payload means that CW and Provider apps can see that its set and do there thing from there

app/jobs/nsm/schedule_delete_reviewed_docs.rb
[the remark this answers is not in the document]
robinsk1: @ivanELEC  i originally had it as `last_updated_at`- i'll change it back to that, that seems like best fit/trade-off

app/jobs/nsm/delete_reviewed_claim_docs.rb
ivanELEC: sorry let me be more specific - currently gdpr_documents_deleted is in the application field, which is a JSON payload. I'm suggesting putting it as a field in the top level of the submission insted of the in the application JSON. The services would still be able to retrieve this via the API

ivanELEC: The main thing that is in conflict with my suggestion is that the top level is reserved for fields that are relevant for all submission types and *technically* this field is currently only relevant for crm7s

app/jobs/nsm/schedule_delete_reviewed_docs.rb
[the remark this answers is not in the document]
ivanELEC: yeah I'd agree - if we end up getting requirements that conflict with this, we can update accordingly - thought I doubt we would as once they hit that final state, I can't see caseworkers caring about what happens with the submission after

app/jobs/nsm/delete_reviewed_claim_docs.rb
ivanELEC: resolved because the way metrics are designed is the issue

app/jobs/nsm/schedule_delete_reviewed_docs.rb
ivanELEC: this is a nit so resolve if you want

app/jobs/nsm/delete_reviewed_claim_docs.rb
```diff
@@ -36,15 +36,13 @@ def perform(submission_id)
 
       submission.with_lock do
         submission.current_version += 1
-        submission.last_updated_at = now
         submission.caseworker_history_events << build_delete_supporting_evidence_event(now, submission.current_version)
         submission.save!
```
elken, line 40: Do we want to do `save!(touch: false)` here also?

robinsk1: if the only footprint we want to leave is the event and the flag, i guess so

elken: Wasn't saying it should or shouldn't, just occurred to me 😄

app/jobs/nsm/schedule_delete_reviewed_docs.rb
```diff
@@ -0,0 +1,24 @@
...
+
+      filtered_claims.each do |claim|
+        Nsm::DeleteReviewedClaimDocs.perform_later(claim.id)
+      end
+    end
+
+    def filtered_claims
+      state = %w[granted part_grant rejected expired]
+
+      Submission.joins("LEFT JOIN application_version latest_version ON latest_version.application_id = application.id AND
```
elken, line 16: Can this be handled instead by ActiveRecord syntax without dipping into SQL?

app/services/file_upload/file_uploader.rb
```diff
@@ -0,0 +1,8 @@
+module FileUpload
+  class FileUploader
+    def destroy(file_path)
```
elken, line 3: Do you not also want to consider file existence in the bucket? I don't know if trying to delete a missing file throws an exception and I don't think it should stop the whole job running.

elken (approved):
Couple of questions but otherwise happy once they're resolved

app/jobs/nsm/schedule_delete_reviewed_docs.rb
robinsk1: Its not really much more readable using AR or Arel with the subquery, going to leave this as it is

PR #374, opened 2025-01-29
Crm457 2386 nsm app store obfuscate filenames for submission version history

elken:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-XXX)


ivanELEC (approved):

app/jobs/nsm/delete_reviewed_claim_docs.rb
```diff
@@ -66,5 +73,25 @@ def build_delete_supporting_evidence_event(now, current_version)
     def file_uploader
       @file_uploader ||= FileUpload::FileUploader.new
     end
+
+    def redact_file_names(application, path)
+      paths = %w[file_path file_name]
+
+      case application
```
robinsk1, line 80: i'm not sure about iterating through the entire application and deleting based on paths, this feels a bit omniscient - we could forget about this later, add an array with a file_name element (file not being an actual s3 file or something within scope of GDPR) it would get blasted irrespective

app/jobs/nsm/delete_reviewed_claim_docs.rb
```diff
@@ -28,6 +28,13 @@ def perform(submission_id)
       docs_to_destroy.flatten.each do |file_path|
         if file_uploader.destroy(file_path)
           Rails.logger.info "Deleted #{file_path} for submission #{submission.id}"
+
+          SubmissionVersion.where(application_id: submission_id).find_each do |submission_version|
+            submission_version.with_lock do
+              submission_version.application = redact_file_names(submission_version.application, file_path)
```
robinsk1, line 34: replacing the entire application payload here feels a bit global in scope - see next comment

spec/jobs/nsm/delete_reviewed_claim_docs_spec.rb
```diff
@@ -67,6 +67,15 @@
           described_class.new.perform(claim.id)
           expect(claim.reload.latest_version.application["gdpr_documents_deleted"]).to be(true)
         end
+
+        it "removes the filenames from the history" do
```
robinsk1, line 71: can we add another test showing this works as expected for RFI - slightly different data structure to supporting_evidence

app/jobs/nsm/delete_reviewed_claim_docs.rb
robinsk1: ok! blast on - can we have an extra test that covers the recursion here just to make sure its doing what we expect 🙏

spec/jobs/nsm/delete_reviewed_claim_docs_spec.rb
elken: @robinsk1 Added a couple more test cases, is that complete enough now?

PR #398, opened 2025-04-24
CRM457-2517: Add new route for creating failed import

ivanELEC:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2517)


app/services/failed_imports/creation_service.rb
```diff
@@ -0,0 +1,15 @@
+module FailedImports
+  class CreationService
+    AlreadyExistsError = Class.new(StandardError)
+    class << self
+      def call(params)
+        raise AlreadyExistsError if FailedImport.find_by(id: params[:id])
+
+        FailedImport.transaction do
+          failed_import = FailedImport.create!(id: params[:id], provider_id: params[:provider_id])
```
elken, line 9: No need to capture this in a variable if it's just being returned

app/controllers/V1/failed_imports_controller.rb
```diff
@@ -0,0 +1,12 @@
+module V1
+  class FailedImportsController < ApplicationController
+    def create
+      failed_import = ::FailedImports::CreationService.call(params)
```
elken, line 4: I'm not convinced there needs to be an entire service for a `find_or_create_by` call and also why we're passing `id` in the first place. It should just be an autogenerated UUID that gets returned when the object is created

ivanELEC: I wanted there to be as much correlation between the app store records and the provider app records - purely to make it easier to perform any data cleaning/deleting processes between the services (as this is a table where the data is on both and needed to be on both)

app/services/failed_imports/creation_service.rb
ivanELEC: fair, I can refactor

app/controllers/V1/failed_imports_controller.rb
ivanELEC: For example, we justify ensuring the ids match on Submissions between the services for a similar reason

elken: Oh you made the change already, my bad

elken (approved):
Looks fine to me 😄

app/controllers/V1/failed_imports_controller.rb
ivanELEC: yeah agreed, am updating that, was referencing the 2nd part of the comment

ivanELEC: sorry I grouped the other comment in with this one too

elken: I also forgot about the UUID bit, I'll make another discussion for it

app/controllers/V1/failed_imports_controller.rb
```diff
@@ -0,0 +1,18 @@
+module V1
+  class FailedImportsController < ApplicationController
+    def create
+      raise AlreadyExistsError if FailedImport.find_by(id: params[:id])
+
+      FailedImport.transaction do
+        failed_import = FailedImport.create!(id: params[:id], provider_id: params[:provider_id])
```
elken, line 7: UUID should be autogenerated not taken as a param, then you don't even need to check if one exists.

(`failed_import` is also redundant, you can just remove it)

elken: I raised the point over on the submit PR that we shouldn't be storing the data there anyway, the app store is the source of truth now.

>  just respond with the status instead of responding with the status AND the created object

I think that's fine practice in general, but in this context yeah there's no extra information in the object so that's fine

elken: If we're storing it in the database like this we don't have to rely on the hack of linking the ID to a specific provider. A `FailedImport` is a unique event of which a provider can have many of, it shouldn't matter if the account is shared or not. It greatly simplifies the work here and as I said over there, the app store should be the source of truth of data going forward.

ivanELEC: that's true - the main benefit of having matching ids is syncing deletes across and as you said, the app store only cares about the event, we aren't storing the details. I'll just have the id auto generate

ivanELEC: updated

ivanELEC: - the only reason we store the provider id is so that we can potentially use analytics to understand who is making these errors and understand if there are interventions we can make on that basis - as you said we are no longer using provider id to generate a filename for the error 
- I mentioned error details in the provider failed import table earlier because that's the mechanism in which the error is persisted (allowing us to not have to "overwrite") 
- you're right in saying we don't need to store the data in the provider application - I wanted to avoid having to create another GET route for the failed import and us having to do an extra transaction with the app store each time an import error happens - however 1). there probably won't be too many of these errors happening after the initial hump of people getting used to the feature and therefore, not too many additional HTTP calls 2). I guess having another route is a small price to pay for SSOT

elken (approved):

PR #422, opened 2025-06-11
[CRM457-2627] send_back expired submissions

robinsk1:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2627)
- [x] common gem changes here https://github.com/ministryofjustice/laa-crime-forms-common/pull/60  

## Notes for reviewer

- Once run, this rake task will unexpire submissions that have been expired after passing their resubmission deadline. 

- Unexpired submissions will be scoped between `2025-5-16` and the day this task is executed. 

- All unexpired submissions will have their resubmission date set to 10 working days in the future

### DIFF
--- a/lib/tasks/CRM457_2627.rake
+++ b/lib/tasks/CRM457_2627.rake
@@ -1,18 +1,30 @@
 # https://dsdmoj.atlassian.net/browse/CRM457-2627
 
+# accepts `start date` and `batch size` params
+# be rake CRM457_2627:send_back_expired['2025-5-16', 10]
+
 namespace :CRM457_2627 do
   desc "send_back_back expired"
-  task send_back_expired: :environment do
-    expired_send_backs.find_each { send_back(_1) }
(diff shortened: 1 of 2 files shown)


lib/tasks/CRM457_2627.rake
```diff
@@ -0,0 +1,62 @@
...
+      updated_at: now,
+      event_type: "send_back",
+      public: false,
+      does_not_constitute_update: false,
+    }
+  end
+
+  def start_date
+    Date.new(2025, 5, 16).beginning_of_day
+  end
```
ivanELEC, line 61: could input the date as an arg e.g. 2025-05-16 and process it that way instead of hardcoding, that way we don't have to make a new PR each time we want to use this

ivanELEC: Not compulsory though, it's likely we'll only use once

ivanELEC (approved):

PR #431, opened 2025-07-15
Create table to persist Payment Request records in system

ivanELEC:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2863)

## Notes for reviewer

- Added table for payment requests
- Added table for CRM7s 
- Added table for CRM8s 
- Added relationships at db level and model level

## Screenshots of changes (if applicable)
ERD for new entities
Screenshot 2025-07-15 at 16 52 16

### DIFF
--- a/app/models/nsm_claim.rb
+++ b/app/models/nsm_claim.rb
@@ -1,5 +1,5 @@
 class NsmClaim < ApplicationRecord
   has_many :payment_requests, dependent: :destroy, inverse_of: :payable, as: :payable
   has_one :assigned_counsel_claim, dependent: :destroy
-  has_one :submission
+  has_one :submission, dependent: :destroy
 end
--- a/app/models/nsm_claim.rb
+++ b/app/models/nsm_claim.rb
@@ -1,0 +1,4 @@
+class NsmClaim < ApplicationRecord
+  has_many :payment_requests, as: :payable
+  has_one :assigned_counsel_claim
+end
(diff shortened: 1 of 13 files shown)


app/models/nsm_claim.rb
```diff
@@ -0,0 +1,4 @@
+class NsmClaim < ApplicationRecord
+  has_many :payment_requests, as: :payable
```
elken, line 2: Are they payable? They're _requests_ to be paid, I think leaving the relationship as-is is fine

elken: Also need to add delete/update cascade behavior for the relationships

ivanELEC: Regarding "payable" it was the most generic naming I could think of for something that needed payments - payment_requestable feels like it doesn't make sense. Also, the alias isn't in there for neatness, I need it to make the payment_request associations polymorphic as it can apply to many different entities (like we have for supportinf documents) 

Invoiceable might be better? but then that might intefere with a future where we link with CCMS and actually send invoices (if we ever have to store those invoices separately to the payment request)

elken: That's fair, it's not a huge gripe. Makes enough semantic sense to just be payable

elken (approved):
Approved assuming you've testing the `dependent` relationship

PR #440, opened 2025-07-29
CRM457-2691: Link Payment Request

ivanELEC:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2691)


app/helpers/laa_reference_helper.rb
```diff
@@ -0,0 +1,30 @@
...
+        random_reference = "LAA-#{SecureRandom.alphanumeric(6)}"
+        # :nocov:
+        break random_reference unless reference_already_exists?(random_reference)
+        # :nocov:
+      end
+    end
+  end
+
+  def reference_already_exists?(laa_reference)
+    return true if CLAIM_CLASSES.any? { _1.exists?(laa_reference: laa_reference) }
```
robinsk1, line 16: Not sure how efficient this is esp with big tables as its loading table then doing the check in ruby. Should be faster doing this in active record

app/models/nsm_claim.rb
```diff
@@ -2,4 +2,6 @@ class NsmClaim < ApplicationRecord
   has_many :payment_requests, dependent: :destroy, inverse_of: :payable, as: :payable
   has_one :assigned_counsel_claim, dependent: :destroy
   has_one :submission, dependent: :destroy
+
+  validates :laa_reference, presence: true
```
robinsk1, line 6: is `laa_reference` here non-unique? we can have multiple claims for same `laa_reference` ?

app/helpers/laa_reference_helper.rb
```diff
@@ -0,0 +1,30 @@
+module LaaReferenceHelper
```
robinsk1, line 1: would scoping/renaming this to `payment requests` indicate intent better?  - maybe we only need to include this within payment_requests_controller

app/helpers/laa_reference_helper.rb
```diff
@@ -0,0 +1,30 @@
...
+  CLAIM_CLASSES = [NsmClaim, AssignedCounselClaim].freeze
+  def generate_laa_reference
+    ActiveRecord::Base.transaction do
+      CLAIM_CLASSES.each(&:lock)
+      loop do
+        random_reference = "LAA-#{SecureRandom.alphanumeric(6)}"
+        # :nocov:
```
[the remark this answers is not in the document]
ivanELEC: yeah guess I can use a timeout or something to test that if the alphanumeric method returns an existing reference, it doesn't exit the loop? My reasoning for not testing this specific section is that the reference_already_exists? method was tested already - so testing this specific block of code just feels like testing loop, unless and break work as expected which felt like testing Ruby and not our specific logic. But I think it's one of those things that could lead to Very Bad Stuff happening if it goes wrong so should test

robinsk1 (commented):
looks good 😄 - just some minor naming stuff feedback that can be ignored if you feel it doesn't add anything

app/helpers/laa_reference_helper.rb
ivanELEC: So this functionality won't just be for payment requests. We have [a ticket in place](https://dsdmoj.atlassian.net/browse/CRM457-2704) - which I might actually bring into the sprint and do next which moves ALL laa reference generation to the app store - meaning when CRM4s and CRM7s get submitted, an laa reference is created at that point (instead of when the application is being drafted). We found that the provider doesn't need to know the laa reference before it's submitted. This means we are only checking one db when trying to find out if an laa reference already exists or not

app/models/nsm_claim.rb
ivanELEC: Yeah, it should be unique, the logic for it's uniqueness is nuanced. For now: 
all laa references in NsmClaim, AssignedCounselClaim and Submission(Version) combined should be unique because there are currently no scenarios where we're creating a payment for an existing digital submission.

 But in the future, we might have functionality to link an existing Submission to an NsmClaim to "auto-create" the payment request. Which means laa references in NsmClaim and Submission won't necessarily have uniqueness 
 
This will be handled in this ticket: https://dsdmoj.atlassian.net/browse/CRM457-2704

app/helpers/laa_reference_helper.rb
ivanELEC: aye I think you're right there

ivanELEC: the any? here is just looping through the (currently 2) claim classes to perform the exists? method that many times

robinsk1 (approved):

PR #442, opened 2025-07-31
CRM457-2694 assigned_counsel_claim update endpoint

robinsk1:
## Description of change

PATCH endpoint for `/assigned_counsel_claim/{object.id}` allowing update of `counsel_office_code`

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2694)

### DIFF
--- a/app/controllers/V1/assigned_counsel_claims_controller.rb
+++ b/app/controllers/V1/assigned_counsel_claims_controller.rb
@@ -5,6 +5,7 @@ module V1
       render json: assigned_counsel_claim, status: :created
     rescue ActiveRecord::RecordInvalid, ActionController::ParameterMissing => e
       render json: { errors: e.message }, status: :unprocessable_entity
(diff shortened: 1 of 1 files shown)


app/controllers/V1/assigned_counsel_claims_controller.rb
```diff
@@ -0,0 +1,22 @@
+module V1
+  class AssignedCounselClaimsController < ApplicationController
+    def update
+      assigned_counsel_claim.update!(allowed_params)
+      render json: assigned_counsel_claim, status: :created
+    rescue ActiveRecord::RecordInvalid, ActionController::ParameterMissing => e
+      render json: { errors: e.message }, status: :unprocessable_entity
```
ivanELEC, line 7: this isn't explicitly in the ticket but could we use the report_error method here so that this gets logged and reported in Sentry

PR #448, opened 2025-08-11
DO NOT MERGE: CRM457-2704: Remove LAA Reference

ivanELEC:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2704)

## Notes for reviewer

- [x] Use generate_laa_reference method to generate a new unique laa ref when creating a new submission
- [x] Send submission email from app store
- [x] Update tests

Additional note: I am not sure if the way I am bringing in presenters here is the right way to go necessarily for us when unifying into the app store. The way presenters are in the provider app are quite different to the way they are in the caseworker app, I've taken bits and pieces from each to bring in the bare minimum of what we need for the PR, a spike with more time and collaboration allocated to it is needed for when we make the app store db the single source of truth for drafts

### DIFF
--- a/app/presenters/V1/nsm/claim_presenter.rb
+++ b/app/presenters/V1/nsm/claim_presenter.rb
@@ -10,15 +10,15 @@ module V1
       end
 
       def work_items_for_calculation
-        @application["work_items"].map { WorkItemCosts.new(_1, @application).data_for_calculation }
+        @application["work_items"].map { WorkItemCostsPresenter.new(_1, @application).data_for_calculation }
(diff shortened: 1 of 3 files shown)


app/errors/no_laa_reference_error.rb
```diff
@@ -0,0 +1,2 @@
+class NoLaaReferenceError < StandardError
```
elken, line 1: We should include a useful message

robinsk1 (approved):

PR #554, opened 2026-03-17
CRM457-2841: Update Work Item Uplift Rounding Logic

ivanELEC:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2841)

## Notes for reviewer

1. Applies v13.0.0 upgrade to Laa Crime Forms Common so new calculation methods for work item totals can be used
2. Updates any tests where calculation totals change 

Proof of local tests succeeding: 
image


robinsk1 (dismissed):

Gemfile.lock
```diff
@@ -190,7 +189,7 @@ GEM
       rdoc (>= 4.0.0)
       reline (>= 0.4.2)
     jmespath (1.6.2)
-    json (2.19.3)
+    json (2.18.1)
```
robinsk1: odd we've gone down?

robinsk1 (dismissed):

Gemfile.lock
ivanELEC: yeah that is odd, I tried regenerating the gemfile from scratch locally and it changes a bunch of these versions so I think the gemfile.lock is generally out of sync. For simplicity, will keep this the same so that this change is independant and then do another PR to try and get through some gem updates

ivanELEC: oh actually looks like that version has vulns so will sort that out now

cmrgov (dismissed):

PR #600, opened 2026-04-29
CRM457-3031: Add composite application search index

cmrgov:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-3031)

Adds a composite index on `application(application_type, last_updated_at)` to support Prior Authority dashboard searches ordered by latest update.

This PR is intentionally index-only so it can be rolled out first. The `searches` view rewrite is split into a separate follow-up PR for next week.

## Notes for reviewer

The new index is created concurrently and is rollbackable.

This index gives Postgres an ordered access path for dashboard queries such as:

```sql
WHERE application_type = 'crm4'
ORDER BY last_updated_at DESC
LIMIT ...
```

Local benchmarking showed the largest improvement comes when this index is combined with the follow-up `searches` view rewrite. This PR lays down the index first so we can deploy it separately and observe safely.

## Screenshots of changes (if applicable)

N/A

Run the migration and rollback locally:

```sh
RAILS_ENV=test bundle exec rails db:migrate
RAILS_ENV=test bundle exec rails db:rollback STEP=1
RAILS_ENV=test bundle exec rails db:migrate
```

### DIFF
--- a/db/migrate/20260429090951_add_index_to_application_on_application_type_and_last_updated_at.rb
+++ b/db/migrate/20260429090951_add_index_to_application_on_application_type_and_last_updated_at.rb
@@ -1,0 +1,22 @@
+class AddIndexToApplicationOnApplicationTypeAndLastUpdatedAt < ActiveRecord::Migration[8.1]
+  disable_ddl_transaction!
+
+  INDEX_NAME = "idx_application_on_type_last_updated_at"
+
+  def up
+    return if index_exists?(:application, %i[application_type last_updated_at], name: INDEX_NAME)
+
+    add_index :application,
+              %i[application_type last_updated_at],
+              name: INDEX_NAME,
+              algorithm: :concurrently
+  end
+
+  def down
+    return unless index_exists?(:application, name: INDEX_NAME)
+
+    remove_index :application,
+                 name: INDEX_NAME,
+                 algorithm: :concurrently
+  end
+end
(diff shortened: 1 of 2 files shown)


robinsk1 (commented):
was there any performance improvements running the query targeted for optimisation after this index was created given the variegated data that you can create via the update bulk dummy data stuff?

ivanELEC (dismissed):

ivanELEC:
Approved but as @robinsk1 pointed out, might be good to test this with dummy data before merging

cmrgov:
@robinsk1 @ivanELEC - good flag.

This composite index doesn't appear to solve our problem as far as I can see. When we query the searches view, it's skipping usage of this index on the base table. I'm doing a deeper investigation and closing this for now.

robinsk1 (commented):
Great stuff! Can we rollback the index change and re-run that view CTE change to see if thats where the improvement lies? Is that something you've already tested

cmrgov:
> Great stuff! Can we rollback the index change and re-run that view CTE change to see if thats where the improvement lies? Is that something you've already tested

The view rewrite alone was a significant improvement for the list query, reducing the reviewed 32-account list query from ~265ms to ~36.5ms. The composite index then gave the larger final improvement, taking it down to ~1.3ms. So the improvement comes from both

For counts, v10 without the index was ~212.8ms versus v09 at ~269ms, and v10 with the index was ~177ms. So both also help counts, but optimising this further remains the more complex problem to solve.

ivanELEC (approved):

PR #621, opened 2026-06-01
CRM457-3056: Make Views More Efficient: Caseworker Dashboard Views

ivanELEC:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-3056)

## Notes for reviewer
Makes submissions_by_date more efficient by: 

- Use WHERE to filter by status and pending to reduce number of rows before aggregating with GROUP BY
- Use JOIN instead of LEFT-JOIN as application will always exist
- Remove counted_values subquery - not needed
- Do expression for total column inline

Could not identify a way to make submission_by_services more efficient than it already is


ivanELEC:
Testing Locally with:
 ~100k CRM4 applications across 200 providers, max 3 versions, default version mix

ivanELEC:
**Test 1 - `SELECT COUNT(*) FROM submissions_by_date`**
Before change: avg execution time - 87ms, execution cost - 5479.32
After change: avg execution time - 92ms, execution cost - 4031.23
5ms increase in execution time, 1448.09 decrease in cost

**Test 2 - `SELECT * FROM submissions_by_date`**
Before change: avg execution time - 251ms, execution cost - 10478.44
After change: avg excution time - 189ms, execution cost - 4026.81
62ms decrease in execution time, 6451.63 decrease in cost

**Test 3 - `SELECT * FROM submissions_by_date WHERE event_on >= CURRENT_DATE - INTERVAL '30 days'`**
Before change: avg execution time -  178ms, execution cost - 4617.98
After change: avg execution time - 173ms, execution cost - 3671.14
5ms decrease in execution time, 946.84 decrease in cost

cmrgov (changes requested):
@ivanELEC I think `schema.rb` is out of sync. It says `2026_06_01_150127`, but that migration was deleted from this branch.

Now `main` has `20260602100000_update_processing_times_to_version_4.rb`, so after rebasing/migrating I’d expect the schema version to be `2026_06_02_100000`.

Might just need a rebase and `rails db:migrate`

db/views/submissions_by_date_v06.sql
```diff
@@ -0,0 +1,12 @@
+SELECT
+  av.created_at::date AS event_on,
+  a.application_type,
+  COUNT(*) FILTER (WHERE av.application ->> 'status' = 'submitted')     AS submission,
+  COUNT(*) FILTER (WHERE av.application ->> 'status' = 'provider_updated') AS resubmission,
+  COUNT(*)                                                                AS total
```
robinsk1, line 6: the 3 counts here look like the performance bottleneck - i guess we see how much this change makes and go from there

cmrgov (approved):

PR #648, opened 2026-06-26
CRM457-3129: Dev: Store Digital Submission Claim Type on Payable Claim Instead of Payment Request

ivanELEC:
[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-3129)

## Notes for reviewer

- Add nsm_claim_type to the nsm_claim model with model validation
- Include nsm_claim_type in params for creation service

### DIFF
--- a/app/models/nsm_claim.rb
+++ b/app/models/nsm_claim.rb
@@ -23,4 +23,5 @@ class NsmClaim < PayableClaim
   validates :stage_code, format: { with: /\A\bPROG\b|\bPROM\b\z/ }
   validates :court_attendances, :no_of_defendants, numericality: { only_integer: true }
   validates :original_submission_date, presence: true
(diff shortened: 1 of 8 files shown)


robinsk1 (changes requested):
I don't think we should be introducing another field here.

The underlying issue is that we're currently using request_type incorrectly. At the moment it isn't normalised—we're conflating claim types (e.g. non_standard_mag, assigned_counsel) with request types (e.g. supplemental) by storing values such as non_standard_mag_supplemental in the request_type attribute. As a result, we've ended up with a lot of claim type checks throughout the codebase just to derive the actual claim type.

Rather than adding another attribute, I think we should refactor this so the distinction between claim type and request type is explicit. Introducing a new field would only add another layer of complexity and muddle things more.

For example, `breach of injunction` fits as a request type within a non_standard_mag claim. We don't need to change the underlying data model—we just need to use it correctly by keeping claim type and request type as separate concepts.

PR #656, opened 2026-07-03
Use correct column name for nsm_payments view

ivanELEC:
## Description of change

No ticket - this is to fix migration issues with nsm_payments: 
Upgrade to version 2 isn't taking effect, this is likely because the first version references a column name that has since been changed (date_received --> date_claim_assessed)
This ticket adjusts version 1 of the view to not have the date_received column at all to prevent migration issues `20260618111231_update_nsm_payments_to_version_2
`


db/schema.rb
```diff
@@ -299,11 +299,11 @@
       payment_requests.claimed_waiting_cost,
       payment_requests.claimed_total,
       payment_requests.allowed_total,
-      payment_requests.date_claim_assessed,
+      payment_requests.date_claim_assessed AS date_received,
```
RobBoothMOJ, line 302: date_received is still present in the 'autogrant_events' view though sourced from date_claim_assessed rather than date_received?

ivanELEC: autogrant_events links to the application_version table, not the payment_requests table

RobBoothMOJ (approved):
LGTM

